PR #1, opened 2019-09-24
Add README section on local setup

fsteeg:
Run local web server for data-include support

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,16 @@ for editor metadata).
 
 This specification is developed using [ReSpec](https://github.com/w3c/respec) and help about
 its markup can be found in its [user guide](https://github.com/w3c/respec/wiki/ReSpec-Editor's-Guide).
+
+Local setup
+-----------
+
+To edit and preview the `index.html` file locally, clone the repo and change into the `specs` directory:
+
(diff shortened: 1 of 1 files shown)


wetneb:
Thanks! With my browser (Firefox 68) it seems that `data-include` works even with `file://`, but it might not be true for everyone.

fsteeg:
> With my browser (Firefox 68) it seems that data-include works even with `file://`

Interesting, I also have Firefox 68 (though it also says 'for Linux Mint' in the info screen), and I got ReSpec errors for the data-includes.

wetneb:
Ok. Also I am not sure about this `data-include` business, perhaps it is not that useful to include the source of the JSON Schemas in the specs (linking to them could be enough). I just wanted to play with this nice feature, but it does not mean we should necessarily have it :)

fsteeg:
> perhaps it is not that useful to include the source of the JSON Schemas in the specs (linking to them could be enough)

I like the inclusion because we get syntax coloring and consistent display right in the document, while with links, a JSON plugin would have to be installed in the browser. Inclusion might also be useful for other JSON snippets in the spec (like complex reconciliation requests including properties etc.), which could be edited in separate files, with a nice editor supporting JSON, and sent as data with curl to validate them (even automated).

PR #4, opened 2019-09-25
Fix response schema

acka47:
The response schema would not validate responses from the lobid-gnd endpoint ([example](https://lobid.org/gnd/reconcile/?queries=%7B%22q1%22%3A%7B%22query%22%3A%22Twain%2C+Mark%22%7D%7D)). This PR changes the root type of the response schema to `object`. It also adds the optional `match` property to the schema.

BTW, looking at the [current documentation]/(https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation-Service-API#query-response), we should make all properties `required`, i.e. also add `type` and `match`.

### DIFF
--- a/schemas/reconciliation-response.json
+++ b/schemas/reconciliation-response.json
@@ -1,7 +1,7 @@
 {
     "$schema": "http://json-schema.org/schema#",
     "$id": "https://reconciliation-api.github.io/specs/schemas/reconciliation-response.json",
-    "type": "array",
+    "type": "object",
     "description": "This schema validates lists of reconciliation candidates",
     "items": {
         "type": "object",
--- a/schemas/reconciliation-response.json
+++ b/schemas/reconciliation-response.json
@@ -18,6 +18,10 @@
                 "type": "number",
                 "description": "Number indicating how likely it is that the candidate matches the query"
             },
+            "match": {
+              "type": "boolean",
+              "description": "Boolean value indicating whether the candiate is a certain match or not."
+            },
             "type": {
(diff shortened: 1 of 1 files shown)


acka47:
I accidentally pushed the changes to master from another branch. @wetneb, you will probably take a look anyway.

wetneb:
Thanks! This schema is meant to validate the response to an individual query. In your example this would be this part of the JSON returned by lobid:

```
[

    {
        "id": "118624822",
        "name": "Twain, Mark",
        "score": 80.17445,
        "match": true,
        "type": [
            {
                "id": "AuthorityResource",
                "name": "Normdatenressource"
            },
            {
                "id": "DifferentiatedPerson",
                "name": "Individualisierte Person"
            }
        ]
    },
    {
        "id": "1045623490",
        "name": "Bezirkszentralbibliothek Mark Twain. Schreibwerkstatt",
        "score": 74.66594,
        "match": false,
        "type": [
            {
                "id": "AuthorityResource",
                "name": "Normdatenressource"
            },
            {
                "id": "CorporateBody",
                "name": "Körperschaft"
            }
        ]
    },
    ...
]
```

Note that this excludes the `{"q0": ... , "q1": ...}` part.

I agree it would also be good to have a schema for the full response. I thought it could be useful to isolate the schema for a single response since it could then be used for both individual and multiple query modes (via `$ref`). But if we decide not to document the individual query mode at all, it would make sense to make it in one single big schema too.

Thanks for adding the `match` parameter. Happy with marking these as required indeed!

Issue #5, opened 2019-09-25
reconciliation-response schema is broken


acka47:
For example with a response `reconciliation-response.json` like 

```json
{
   "q1":{
      "result":"bla"
   }
}
```

this is also validated as correct:

`$ ajv -s schemas/reconciliation-response.json -d reconciliation-response.json 
reconciliation-response.json valid`

wetneb:
That is because it was not meant to be used for the full response, only the list of candidates for a single query, as I tried to document in its description :)

wetneb:
That is because it was not meant to be used for the full response, only the list of candidates for a single query, as I tried to document in its description :)

acka47:
Sorry, then my other commits also don't make sense. Next time I will read the description first. ;-) However, wouldn't it make sense to have a schema for the full response. I think, we just have to take into accout the arbitrary keys (in the example `q1`) which should be easily done with `patternProperties`, see https://stackoverflow.com/questions/16222633/how-would-you-design-json-schema-for-an-arbitrary-key.

wetneb:
yes, see my comments on the PR

acka47:
You write at :

> I thought it could be useful to isolate the schema for a single response since it could then be used for both individual and multiple query modes (via $ref). But if we decide not to document the individual query mode at all, it would make sense to make it in one single big schema too.

As the single query mode is marked as deprecated in the documentation, I would go on and adjust the schema to validate the whole response. Then it would also fit with the text in the draft spec, see e.g. https://reconciliation-api.github.io/specs/#reconciliation-response-schema Ok?

wetneb:
I think it is consistent with the text at the moment, since reconciliation response is defined as "a set of reconciliation candidates".

But sure, happy with that plan. (So the text would need to be updated)

acka47:
Ok, will do that. (I meant the bit in the spec: "The reconciliation response schema can be used to validate the JSON serialization of any reconciliation response.")

wetneb:
Yes, if you click on "reconciliation response" you get to the definition ;)

acka47:
> Also, for consistency, if we have a single schema for multiple responses, then it would probably also make sense to have a single schema for multiple queries.

Yes, I made a PR for this with #9.

PR #8, opened 2019-09-25
Make response schema validate whole response

acka47:
... and add `type` and `match` as required.
Fixes https://github.com/reconciliation-api/specs/issues/5.

See also https://github.com/reconciliation-api/specs/pull/4#issuecomment-534996592


acka47:
You are referring to https://github.com/reconciliation-api/specs/blob/ed2fc029efc3d09cfb79110b421078e6ed1ce111/index.html#L257
I am not sure how to update it. What do you think about something like "A reconciliation response is a set of reconciliation candidates for each query string provided"?

wetneb:
Yes! It might need more than a single sentence to be completely clear, using intermediate definitions, but we can tweak that later on.

acka47:
I just chatted with @fsteeg and we agreed upon rather using the term "result" than "response" ("reconciliation result" and "reconciliation result batch") because "response" is regularly used for the server response to a HTTP request. As the spec uses "response" in both meanings it might lead to confusion

https://github.com/reconciliation-api/specs/blob/35c0f1e47d61dfc51af5559f698f3f0c0884c0fb/index.html#L202

What do you think, @wetneb ?
I won'tchange it when used in the context of data extension ("Data Extension Response"), though as the common meaning applies here.

acka47:
With 918d404f7673b4b05543e1118d84c7ecc8e69c8c, I updated the terminology as written in  https://github.com/reconciliation-api/specs/pull/8#issuecomment-535493677, added a defnition for result batch and updated references and descriptions of the schema. Please review, @wetneb .

acka47:
Great, I added myself as editor. Merging.

PR #9, opened 2019-09-25
Validate whole query

acka47:
Corresponding adjustment to #8 for the query schema.


schemas/reconciliation-query.json
```diff
@@ -1,77 +1,93 @@
...
-                            },
-                            {
-                                "type": "number"
-                            },
-                            {
-                                "type": "boolean"
+    "$schema":"http://json-schema.org/schema#",
+    "$id":"https://reconciliation-api.github.io/specs/schemas/reconciliation-query.json",
+    "type":"object",
+    "description":"This schema validates a single reconciliation query",
```
wetneb: The description should also be updated, I think

wetneb:
I would also update the text: to say that the schema validates a query batch.

schemas/reconciliation-query.json
wetneb: And perhaps the schema should be renamed to something like `reconciliation-query-batch.json` to avoid ambiguities.

acka47: For now I won't be renaming the file. If there is only one type of query ("Multiple Query Mode" in the OR docs), I don't think we have to be more specific. IMO, it makes only sense if there also is another kind of query.

Regarding the description, I would just change it to "This schema validates a reconciliation query". Similar to the above, if there is only one query mode, this is totally fine I think. Ok?

acka47:
I wasn't aware my changes in the schemas would require changes in the spec otherwise I would have approached it differently (or just left it). However, do we also have to adjust this:

https://github.com/reconciliation-api/specs/blob/ed2fc029efc3d09cfb79110b421078e6ed1ce111/index.html#L229-L238

or only

https://github.com/reconciliation-api/specs/blob/ed2fc029efc3d09cfb79110b421078e6ed1ce111/index.html#L240

?

schemas/reconciliation-query.json
wetneb: I don't know about you, but I would say that `{"query":"foo","type":"bar"}` is a JSON object that represents a "reconciliation query" (or would you use another term?)

The original docs do say "Multiple Query Mode": so it does mean that an object `{"q0":... ,"q1":...}` contains multiple queries. So an object `{"q0":..., "q1",…}` is not a query itself, it is a group of queries, or query batch, or query set…

wetneb:
I don't think we can just refer to the JSON schemas to define a notion:
* JSON schemas are not very readable for the user
* They do not specify everything (for instance, they do not specify that the strings appearing as type ids in queries should be valid type identifiers, or that an entity should have the same type when returned as a reconciliation candidate and when returned as suggest candidate).

As a service implementer, would you be happy if the specs just gave you a set of JSON Schemas, saying "a reconciliation service is something that takes a JSON blob satisfying this schema and returns a JSON blob satisfying that schema, period"?

The way I would update the text would be to introduce a new notion (leaving the current definition of "reconciliation query" as it is)

> A <dfn>query batch</dfn> is a set of <a>reconciliation queries</a> indexed by string identifiers.

And then update the description of the schema to say that it serializes query batches, not queries.
Wouldn't that be clearer?

acka47:
I totally agree with you that the schema is not the spec and I did not want to suggest that we only need the schema. I must confess, I did not read the complete spec before trying out and committing to the JSON schemas and – as @fsteeg was doing the detail work implementing it at lobid-gnd – I even hadn't been very familiar with the reconciliation API when starting this.

One important takeaway from this: We assumed that the terms "query"/"response" would refer to the HTTP `GET`/`POST` requests and that the schemas would validate the respective payload. With this assumption, I started trying out the schemas on the lobid-gnd reconciliation endpoint. I believe that others will probably think so as well so we should take this viewpoint into account when further developing the spec.

In any way, by now I am way more familiar with the API. :-)

acka47:
> The way I would update the text would be to introduce a new notion (leaving the current definition of "reconciliation query" as it is)
>
> `A <dfn>query batch</dfn> is a set of <a>reconciliation queries</a> indexed by string identifiers.`
>
> And then update the description of the schema to say that it serializes query batches, not queries.
Wouldn't that be clearer?

From the point of view outlined it in the previous comment, I think the distinction between "query" and "query batch" adds unnecessary additional complexity to the spec. If the only way to query a service is with a batch query (even if I only send one single query) and if the service always responds with a "batch response" I would rather keep it simple. I would add this distinction if we wanted to allow single queries without the batch format in the future.

wetneb:
Yes I totally understand that you want to be able to validate the input and output of the service as a whole with a JSON schema, and that is why I am happy with providing schemas for these.

However, I really think it is worth using terms that are unambiguous for everyone, so I insist (sorry!) that we use some form of plural to designate objects of the form `{"q0":…, "q1":…}`. For all the implementations and specifications I am aware of, the term "query" refers to an object of the form `{"query":"foo","type":"bar"}`, so it will be confusing if we use this term to mean something else. Here are a few examples:
* In the original specifications: https://github.com/OpenRefine/OpenRefine/wiki/Reconciliation-Service-API
* In lobid's implementation: https://github.com/hbz/lobid-gnd/blob/master/app/controllers/Reconcile.java#L294
* In OpenRefine's implementation: https://github.com/OpenRefine/OpenRefine/blob/master/main/src/com/google/refine/model/recon/StandardReconConfig.java#L315
* In Conciliatior: https://github.com/codeforkjeff/conciliator/blob/master/src/main/java/com/codefork/refine/SearchQuery.java#L12
* and many more examples, see the sources of the reconciliation services listed at https://reconciliation-api.github.io/testbench/

Moreover, the API itself says (and this is not something we can change easily) that objects of the form `{"q0":…, "q1":…}` are sent to reconciliation services via the `queries=` GET parameter. Note the quer**ies**, not quer**y**.

So I think there is an overwhelming consensus for this nomenclature. I do not think this distinction adds unnecessary complexity to the specs: specifications must be clean and precise. It is not overly bureaucratic to make a distinction between queries and groups of queries: our readers are intelligent and totally capable of understanding the difference, I think.

acka47:
Ok, I tried to adjust the spec as you outlined. See the latest commit. If this is ok, I would do the same for #8.

I wonder, whether we should add something to the description of the schema to make it clearer, like:

> The reconciliation query batch schema can be used to validate the JSON serialization of any reconciliation query batch. i.e. the payload of a `GET`/`POST` to the reconciliation endpoint.

acka47:
Great, thank you. I have learned a lot here. 

I adjusted the description both in the schema and in the spec and will merge this and then update #8.

PR #13, opened 2019-10-01
Expand data extension section

fsteeg:
Service metadata, settings, HTTP requests, examples.

I will add details on property proposals and responses next.

@wetneb If the direction here is good feel free to merge at this point and I'll continue in master.


fsteeg:
> [...] making them available for testing? See the master branch for a suggestion of a new structure

Very cool, great to have a test setup for validating the examples!

I've hooked the data extension examples into the test setup, and tweaked the response schema.

@acka47: Could you take a look at the schema and the tests?

acka47:
This test setup is easy and great, @wetneb. I will definitely reuse it in other standardiziation efforts where we are using JSON schema.

PR #16, opened 2019-10-17
Introduce UML diagram for core data model

wetneb:
This is an attempt to represent the data model as an UML class diagram.
To do so, I have represented the relation between entities and their properties and property values using a `Statement` model (similar to a RDF statement). If we go down that route, we will need to adapt the rest of the presentation.

I don't know if this is helpful or not (it could be intimidating for people who are not well versed into UML - like me)…

uml_diagram


wetneb:
It does not really help and makes things more confusing, I would say.

Issue #32, opened 2020-01-11
Submitting queries as JSON without URL-encoding them


wetneb:
It is more natural in a REST API to submit a JSON payload as a POST query with `Content-Type: application/json` rather than to URL-encode it in a `queries=` parameter as we are currently doing.

We could introduce this change in new version, and perhaps keep URL-encoding for the GET method, since it could still be useful for interactive testing, although the testbench should hopefully decrease the need for this in the future.

Changing this would make it easier to expose OpenAPI profiles (#17) for reconciliation services.

osma:
+1 for making the payload `application/json` which is the normal convention.

I just submitted #33 and I suggest that both changes could be implemented as the same time, as both are breaking changes (though in both cases, a service endpoint could be written to support both the old and the new style).

thadguidry:
> 
> It is more natural in a REST API to submit a JSON payload as a POST query with `Content-Type: application/json` rather than to URL-encode it in a `queries=` parameter as we are currently doing.
> 
👍  agree as well

Issue #33, opened 2020-01-11
Avoid arbitrary keys in query and result batches


osma:
Currently a query batch is a JSON object where the keys can be anything, e.g. (from the [current spec, Example 3](https://reconciliation-api.github.io/specs/latest/#example-3))

```json
{
  "q1": {
    "query": "Hans-Eberhard Urbaniak"
  },
  "q2": {
    "query": "Ernst Schwanhold"
  }
}
```

The arbitrary key aspect makes it difficult to define a clear JSON Schema data model, which could be used within an OpenAPI spec to validate the data. It is possible to write a schema (the spec [has one already](https://reconciliation-api.github.io/specs/latest/#reconciliation-query-batch-schema)), but this requires using a regex `"^.*$"` for the key part, which I think isn't very elegant and could potentially make it difficult to detect genuine problems. See [this comment](https://github.com/reconciliation-api/specs/issues/17#issuecomment-546394670) where the issue was mentioned.

I suggest changing the batch definition to a JSON list where each query has a mandatory `qid` field specifying the query ID, like this:

```json
[
  { 
    "qid": "q1",
    "query": "Hans-Eberhard Urbaniak"
  },
  { 
    "qid": "q2",
    "query": "Ernst Schwanhold"
  }
]
```

Likewise, the result batch would switch from an object indexed using query IDs to a list of result objects where each result object has a `qid` field matching the query ID of the corresponding query.

wetneb:
If we are to use an array, can't we just omit the query id altogether and just use the query order? The responses would have to be in the same order.

For instance, ElasticSearch's bulk API relies on the order of queries:
https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

By the way, ElasticSearch's API actually uses line-delimited JSON, not even an array, to submit multiple queries. I wonder if that is something we should consider - perhaps it makes it easier to submit queries in a streaming fashion. But I assume OpenAPI is not going to like that?

osma:
Yes, using the order would work too. I'm OK with that as well - just want to get rid of the arbitrary keys.

Newline delimited JSON (known as NDJSON or JSON Lines) would be an option too, but as you say, it doesn't play well with OpenAPI. I think you can specify that a payload is in that format, but you cannot then do any validation on the content.

[...]

Issue #17, opened 2019-10-25
OpenAPI profile


wetneb:
Citing @osma at https://github.com/NatLibFi/Annif/issues/338

> it would make the implementation easier if there was an OpenAPI (Swagger) spec for the API available

I am not familiar with OpenAPI but it does sound like something that would be worth having!

workergnome:
We've even working on this at the Getty (for our API documentation).  It's can be hard to document, since OpenAPI doesn't really have a way to document the "JSON encoded as a property" structure that the OpenRefine API spec uses.  

https://gist.github.com/workergnome/afe5b74cff8f1b4fb6490667cf6a4886

wetneb:
In my (very limited) understanding of OpenAPI, OpenAPI profiles are designed to describe the behaviour of a particular instance of a service, and not a protocol that many services can implement. The OpenAPI profiles I am aware of all include the full URL of the concrete service described.

If that is correct, how can a spec writer provide an OpenAPI profile to an implementer? Do we have any example of specs which provide an OpenAPI profile?

osma:
That's a valid concern @wetneb. As you say, an OpenAPI spec is generally specific to a particular instance living at a particular server URL.

My understanding of OpenAPI/Swagger is also quite limited and possibly outdated since I've mostly used version 2.0, not the most recent 3.0, but here are some pointers:

The spec file doesn't have to specify an absolute URL path. In OpenAPI 2.0, this is done using [schemes, host and basePath](https://swagger.io/docs/specification/2-0/api-host-and-base-path/); in 3.0, it is instead specified using the [servers array](https://swagger.io/docs/specification/api-host-and-base-path/) which lists one or more base URLs where the service is available. But these don't have to be absolute URLs; they can also be relative to the URL where the spec file is being served on the web. As a concrete example, consider the Annif tool which is implemented using the OpenAPI-spec-driven Connexion toolkit: here, the [OpenAPI spec for the Annif REST API](https://github.com/NatLibFi/Annif/blob/master/annif/swagger/annif.yaml) (still using 2.0) only states the following:

    schemes:
      - http
    basePath: /v1

[...]

PR #37, opened 2020-02-10
Authentication mechanism

### LINKED ISSUE #26: Authentication support
We would like to make it possible for reconciliation services to support authentication by letting clients pass an API key.

For instance, clients could send an API key along with their queries to the service, via a new parameter (such as `api_key`).

Services would be able to announce in their manifest that they support or require API keys, with perhaps extra metadata to make it easier to prompt the user for such keys.

Mailing list thread: https://lists.w3.org/Archives/Public/public-reconciliation/2019Jul/0010.html
OpenRefine issue: https://github.com/OpenRefine/OpenRefine/issues/2015

wetneb:
This is a proposal for the authentication mechanism proposed on the mailing list:
https://lists.w3.org/Archives/Public/public-reconciliation/2019Jul/0010.html

This proposes to use GET parameters, however I am not sure this is a good idea. Since we also use POST requests, this would mean submitting the `api_key` as a GET parameter to a POST request, which is not very natural (and not very easy to do with many HTTP libraries).

Authentication in GET parameters also has the downside that authentication is leaked in logs.

The motivation for GET was that it was easier to debug requests in a web browser, but browser-based testing will not be doable anyway if we move to pure JSON batches submitted by POST (#32).

I would propose to use a header instead, such as `X-API-Key` suggested by Swagger:
https://swagger.io/docs/specification/authentication/api-keys/

Closes #26.

### DIFF
--- a/latest/examples/manifest/invalid/bad-authentication.json
+++ b/latest/examples/manifest/invalid/bad-authentication.json
@@ -1,0 +1,8 @@
+{
+  "name": "OpenCorporates Reconciliation Service",
+  "identifierSpace": "http://rdf.freebase.com/ns/type.object.id",
+  "schemaSpace": "http://rdf.freebase.com/ns/type.object.id",
+  "authentication": {
+     "foo": "bar"
+  }
+}
--- a/latest/examples/manifest/valid/authentication.json
+++ b/latest/examples/manifest/valid/authentication.json
@@ -1,0 +1,14 @@
+{
+  "view": {
+    "url": "http://vivo.med.cornell.edu/individual?uri={{id}}"
+  },
+  "identifierSpace": "http://vivo.med.cornell.edu/individual/",
+  "name": "VIVO Reconciliation Service",
+  "schemaSpace": "http://vivo.med.cornell.edu/individual/",
+  "authentication": {
+    "tokenName": "Client secret",
+    "required": false,
+    "documentation": "https://vivo.med.cornell.edu/reconciliation/authentication",
+    "format": "[A-Z]{4}[0-9]{7}"
+  }
+}
--- a/latest/index.html
+++ b/latest/index.html
@@ -362,7 +362,7 @@ provide when setting up the service;</dd>
            requests.
         </p>
         <p>
-           When an invalid authentication token is supplied in any HTTP request, the service MUST return an HTTP 403 error,
+           When an invalid authentication token is supplied in any HTTP request, the service MUST return an HTTP 401 error,
            even if authentication is not required.
         </p>
       </section>
--- a/latest/index.html
+++ b/latest/index.html
@@ -300,6 +300,8 @@ database are instances of this type.<dd>
           </dd>
           <dt><code>extend</code></dt>
           <dd>A <a>data extension metadata</a>, supplied if the service offers a <a href="#data-extension-service">data extension service</a>.</dd>
+          <dt><code>authentication</code></dt>
+          <dd>An <a>authentication specification</a>, supplied if the service supports <a href="#authentication">authentication</a>.</dd>
         </dl>
       </p>
       <p>For instance, a service could expose the following minimal service manifest:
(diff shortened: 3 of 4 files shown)


thadguidry:
I still think that Authentication can be left out of the Reconciliation-API specs itself, and instead approached with a Recommending or Guideline instead?  Something how Google or Mozilla or Owasp approach this?
It should be up to the Services themselves to determine how critical their "Authorization" needs are in response to "Authentication".
OAuth 2.0 might be more preferred for providers that need to apply limits against particular resources, and since OAuth 2.0 combines both Authentication and Authorization, this becomes somewhat easier.
Bearer tokens (API keys) might be preferred by others for simple identity lookups because they service only 1 resource type.
Basic digest might be fine if someone is just trying to implement simple throttling of users on their own VM.

In other words, I don't think the Reconciliation-API specs should dictate which common API authentication mechanism can be used,  HTTP Basic, API Keys, OAuth 2.0 , but instead educate providers/developers about them and point to external resources for security awareness.

wetneb:
> I still think that Authentication can be left out of the Reconciliation-API specs itself, and instead approached with a Recommending or Guideline instead? Something how Google or Mozilla or Owasp approach this?

In that case I don't see how to add support for authentication in OpenRefine. What would the user experience look like?

thadguidry:
Ah, for OpenRefine, that's a client that's conforming to the Reconciliation-specs.  A client would/should understand how to parse objects from the OpenAPI JSON or YAML returned, in particular a [Security Requirement Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#schema) object and any provided [securitySchemes](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#components-object).  Then with that metadata knowledge a client could offer a UI or Config message to prompt and store the required values needed from the chosen Security Requirement Object.

In other words, I could discover how to generally interact with a Reconciliation service by querying its [OpenAPI root document that describes its capabilities](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#openapi-object)

Is OpenAPI really required?
* No

Is OpenAPI a good idea?
* Absolutely Yes

What does the above do for Reconciliation service providers?
* It puts a bit more work on Service providers to describe the Reconciliation service
* It benefits providers, because there is a machine readable way to describe the service parameters, security, etc.

What does the above do for Reconciliation clients?
* It allows them to easily discover the capabilities of the Reconciliation service and config their client accordingly.

What does the above do for OpenRefine client?
* Makes our job a bit harder :)
* Helps with some of the other issues of how to describe the API for various parts (parameters, scoring, options, etc)
* Allows understanding of Reconciliation service offerings to generate a comfortable UI for the user

There's [implementations ](https://github.com/OAI/OpenAPI-Specification/blob/master/IMPLEMENTATIONS.md) in all major programming languages for OpenAPI.  Some folks may view OpenAPI as unnecessary, but it solves many of the problems of knowing the capabilities and settings of the Reconciliation endpoint for clients like OpenRefine to make the user experience that much easier.  I.E. a bit more pain on both client/server sides, in order to make users happier and reduce their pain.

wetneb:
> In other words, I could discover how to generally interact with a Reconciliation service by querying its OpenAPI root document that describes its capabilities

Ah ok, that's a more precise proposal then. That would mean adding support for [the following authentication mechanisms](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#securitySchemeObject):
* `apiKey`: that is what was discussed earlier and what this PR proposes
* `http`: basic HTTP auth
* `oauth2`: probably quite painful to implement and to use (since users would need to obtain OAuth client ids) but could be useful for hosted reconciliation client;
* `openIdConnect`: potentially a bit easier to use if endpoints support dynamic client registration…

It would be pretty involved to support all this in clients, but why not.

thadguidry:
Is that a "strong" why not, or "casual" ?  :)  I'm not an authority here, just giving my thoughts.  Tony also had some ideas when he and I talked about Swagger and OpenAPI stuff sometime last year, so I just threw up some of what I remembered he said (I was learning more at that time, still am).  You might ping him on his email also, if you need to.

wetneb:
It's a "why not" that depends on what other people think! :) If we are borrowing more and more concepts from OpenAPI, we might want to do a big overhaul of the entire specs to make it fit in that model (#17, #32, #33 and probably many other aspects we haven't thought about). It is a bit of work and likely to create a backwards-incompatible spec in many ways. But it might well be worth doing.

thadguidry:
Ok, yeah my exact thoughts as well...might be worth the pain.

Issue #31, opened 2020-01-11
Give users more control over candidate scoring


wetneb:
As an API user, the scores returned by reconciliation services are generally opaque: I have no idea what scoring method a service uses without diving deep in its source code. I do not know what a certain score means and it is hard to decide on a threshold above which I can trust matches because of that.

There are basically two classes of approaches to solve this problem.

The first one would be to add more query parameters, to let the user say "it is very important that this particular property matches well" or "the type constraint is only indicative" or similar things (#24). By doing this we give more control to the user over the scoring function. The problems I see with this approach are as follows:
* These new configuration options will still be a bit hard to define for the service and to understand for the user: how to measure the importance of a feature if you don't really know what the scoring function is?
* If I reconcile a column with some settings and then figure out that I want to increase the importance of a particular feature, I will have to re-run the entire reconciliation process since the scoring method is still server-side. This costs time and resources. Therefore this workflow does not tie in very well with machine learning approaches to record linkage.

The second one would be to expose more features in reconciliation responses. The reconciliation service could return not just a single score, but also the list of features it used to compute this global score. Something like this:
```json
{
    "result": [
        {
            "id": "Q2345",
            "name": "My item",
            "type": [ ... ],
            "score": 57.89,
            "features": {
                 "name_levenshtein": 89.63,
                 "name_tfidf": 47.82,
                 "type_match_Q5": 12.3,
                 "type_match_Q47": 50.0,
                 "p_P3477_match": 38.0,
                 "pagerank": 374.89,
                 "nb_statements": 56,
            }
        },
        ...
     ]
}
```
As a client, this would make it possible to come up with your own scoring method based on the features returned by the service, and possibly other ones computed locally. Many of these features require global statistics about the entire database (the PageRank of an entity in a graph, the TF-IDF scores of a sequence of words in a given field), so it makes sense to let the service compute them.

The main difficulty with this approach is coming up with *simple* user interfaces to help users create their own scoring functions without having to know too much about machine learning. This could be done by implementing some active learning strategy where the user only has to manually a few reconciliation candidates (strategically picked by the software), to learn a simple candidate classifier on the basis of these features.
On the plus side, this would make the reconciliation API more relevant to existing software projects which implement feature-based record linkage: [recordlinkage (python package)](https://recordlinkage.readthedocs.io/en/latest/), [reclink (R package)](http://fmwww.bc.edu/RePEc/bocode/r/reclink.html), and many others.

[...]

PR #38, opened 2020-02-10
Expose matching features for client-side scoring

wetneb:
This is a fairly simple proposal to expose matching features in reconciliation candidates, for #31.

The idea behind this proposal is that clients would then be able to construct datasets of reconciliation candidates and their features, annotate which candidates are correct, and train some classifier to predict correctness based on the features (and potentially other features computed locally). This relies on the assumption that services return a given feature (designated by its `id`) in many different candidates.

Note that there is no requirement that the global score for a candidate is derived from the individual features in a particular way: services are free to expose features which are not actually used to compute the global score. This seems useful since it lets services expose features which could potentially be useful in some scenarios, but are not used for the main score to keep its computation simple.

The global score is still kept for backwards-compatibility and as it gives a useful baseline for clients to build on (or potentially use it as a feature itself if it isn't already derived from the exposed features).

This JSON syntax is a bit heavy if services want to return hundreds of features for each candidate, but I wanted to avoid using a simple dictionary as it can make implementation in clients harder (#33):
```json
{
   "tfidf": 3498,
   "pagerank": 34.3,
   "type_match": -11.3
}
```

It's obviously quite hard to evaluate what could go wrong without building the corresponding functionality in clients (for instance OpenRefine)… Do you think such an API would work for your use cases?

### DIFF
--- a/latest/examples/manifest/invalid/feature-view.json
+++ b/latest/examples/manifest/invalid/feature-view.json
@@ -1,0 +1,8 @@
+{
+    "name": "VIAF",
+    "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",
+    "schemaSpace": "http://vocab.getty.edu/doc/#The_Getty_Vocabularies_and_LOD",
+    "feature_view": {
+        "url": "https://vocab.getty.edu/recon/features/"
+    }
+}
--- a/latest/examples/manifest/valid/feature-view.json
+++ b/latest/examples/manifest/valid/feature-view.json
@@ -1,4 +1,5 @@
 {
+    "versions": ["0.2"],
     "name": "VIAF",
     "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",
     "schemaSpace": "http://vocab.getty.edu/doc/#The_Getty_Vocabularies_and_LOD",
--- a/latest/examples/manifest/valid/feature-view.json
+++ b/latest/examples/manifest/valid/feature-view.json
@@ -1,0 +1,8 @@
+{
+    "name": "VIAF",
+    "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",
+    "schemaSpace": "http://vocab.getty.edu/doc/#The_Getty_Vocabularies_and_LOD",
+    "feature_view": {
+        "url": "https://vocab.getty.edu/recon/features/{{id}}"
+    }
+}
--- a/latest/examples/reconciliation-candidate/valid/example.json
+++ b/latest/examples/reconciliation-candidate/valid/example.json
@@ -1,0 +1,34 @@
+{
+  "id": "1117582299",
+  "name": "Urbaniak, Hans-Eberhard",
+  "score": 85.71888,
+  "features": [
+    {
+      "id": "name_tfidf",
+      "value": 378.239
+    },
+    {
+      "id": "pagerank",
+      "value": -3.1209
+    },
+    {
+      "id": "type_match",
+      "value": 10.329
+    },
+    {
+      "id": "deprecated",
+      "value": false
+    }
+  ],
+  "match": true,
+  "type": [
+    {
+      "id": "AuthorityResource",
+      "name": "Normdatenressource"
+    },
+    {
+      "id": "DifferentiatedPerson",
+      "name": "Individualisierte Person"
+    }
+  ]
+}
--- a/latest/examples/reconciliation-result-batch/valid/example-full.json
+++ b/latest/examples/reconciliation-result-batch/valid/example-full.json
@@ -6,6 +6,24 @@
         "name": "Urbaniak, Regina",
         "score": 53.015232,
         "match": false,
+        "features": [
+          {
+            "id": "name_tfidf",
+            "value": 378.239
+          },
+          {
+            "id": "pagerank",
+            "value": -3.1209
+          },
+          {
+            "id": "type_match",
+            "value": 10.329
+          },
+          {
+            "id": "deprecated",
+            "value": false
+          }
+        ],
         "type": [
           {
             "id": "AuthorityResource",
@@ -41,6 +59,24 @@
         "id": "123064325",
         "name": "Schwanhold, Ernst",
         "score": 86.43497,
+        "features": [
+          {
+            "id": "name_tfidf",
+            "value": 334.188
+          },
+          {
+            "id": "pagerank",
+            "value": -4.1581
+          },
+          {
+            "id": "type_match",
+            "value": 13.78
+          },
+          {
+            "id": "deprecated",
+            "value": false
+          }
+        ],
         "match": true,
         "type": [
           {
--- a/latest/index.html
+++ b/latest/index.html
@@ -192,11 +192,16 @@ href="https://github.com/OpenRefine/OpenRefine/wiki/Reconciliable-Data-Sources">
            Moreover, for each <a>property</a> it contains a set of associated <a>property values</a>, possibly empty.
         </p>
         <p>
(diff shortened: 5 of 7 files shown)


rybesh:
Is there any additional information that could make these scores more usable to a human making a judgment about reconciliation candidates (i.e. not only for training a classifier)? For example, a URL for an explanation of the feature/score?

I would potentially use something like this to expose the separate label, temporal and spatial scores for candidates returned by the [PeriodO reconciler](https://github.com/periodo/periodo-reconciler#how-reconciliation-works), but the mostly likely use of them would be to give the person using the reconciler a UI for sorting, comparing and selecting candidates based on the scores.

wetneb:
Great point. It would totally make sense to add more metadata to these features - we want them to be as easy to understand as possible for users. It is true that we need to cater for manual workflows.

I am not sure where to add documentation URLs (and other useful metadata) to features: it feels a bit wasteful to add them in each reconciliation candidate as this will take up a lot of bandwidth. Any idea how to expose that?

rybesh:
A `featureIds` dict in the service manifest? A new kind of `suggest` service?

wetneb:
> A `featureIds` dict in the service manifest?

I was imagining that for the Wikidata reconciliation service, I would generate features for each property supplied in the query: they would be indexed by the property id. For instance, if the user supplies `{"pid":"P1234","v":"foo bar"}` then I would like to return features `"P1234_likelihood"`, `"P1234_fuzzymatch"` and the like.

So the domain of feature ids would be open in my case (the set of valid property ids is infinite since I accept property paths too), making it impossible to list them all in the manifest.

> A new kind of suggest service?

Perhaps a dedicated endpoint would do, yes.

Another option, slightly less wasteful than including feature metadata in reconciliation candidates, would be to include it at the root of reconciliation responses (where metadata for a given feature would be given only once even if the feature appears in multiple reconciliation candidates). But a dedicated endpoint could be cleaner, I think.

acka47:
This looks overall good to me.

> I am not sure where to add documentation URLs (and other useful metadata) to features

One approach for linking to documentation is to make OpenRefine payloads JSON-LD by adding a `@context`. This might mean to adjust a lot of other things but I want to bring this up here anyway as now is the right time to think about it. In this case, you could provide something like this where the documentation would be available behind the URI for each property (here e.g. `https://example.org/score`) and type (here e.g. `https://example.org/Name_tfidf`):

```json
{
    "@context":{
        "name":"https://schema.org/name",
        "score":"https://example.org/score",
        "Name_tfidf":"https://example.org/Name_tfidf",
        "Pagerank":"https://example.org/Pagerank"
    },
    "id":"1117582299",
    "name":"Urbaniak, Hans-Eberhard",
    "score":85.71888,
    "features":[
        {
            "@type":"Name_tfidf",
            "value":378.239
        },
        {
            "@type":"Pagerank",
            "value":-3.1209
        }
    ],
    "match":true,
    "type":[
        {
            "id":"AuthorityResource",
            "name":"Normdatenressource"
        },
        {
            "id":"DifferentiatedPerson",
            "name":"Individualisierte Person"
        }
    ]
}
```

wetneb:
Could this `@context` live higher up in the JSON tree: at the level of the reconciliation result batch, instead of being added to every candidate? If so that would potentially remove some redundancy (and would be more or less what I propose above but in a more principled format).

Generally, formatting responses with JSON-LD seems like a sensible move. Should it be done at other places in the API to make it more uniform? I would just like to keep an eye on the size of the payloads: it would be good not to add too much overhead.

acka47:
Re. removing redundancy and limit the size of the payload wen adding a `@context`: We would have to add the context to each candidate but we could also reference an external context (see [spec](https://www.w3.org/TR/json-ld11/#the-context)) and host the context file itself somewhere else so that only one key-value pair would be added to each response, e.g.:

```json
{
    "@context": "https://openrefine.org/context.jsond",
    "id":"1117582299",
    "name":"Urbaniak, Hans-Eberhard",
    "score":85.71888,
    "features":[
        {
            "@type":"Name_tfidf",
            "value":378.239
        },
        {
            "@type":"Pagerank",
            "value":-3.1209
        }
    ],
    "match":true,
    "type":[
        {
            "id":"AuthorityResource",
            "name":"Normdatenressource"
        },
        {
            "id":"DifferentiatedPerson",
            "name":"Individualisierte Person"
        }
    ]
}
```

wetneb:
Ok, so I guess instead of `https://openrefine.org/context.jsonld`, services would use a URL of their own, which would describe their own features. But then, we are back to the problem of infinite feature sets. If your features can be parametrized (as they would be for Wikidata), then you cannot list them all in this context document. So you would need to automatically generate context documents which contain the metadata of the feature ids used in a particular candidate.

If we want to save space and go down the JSON-LD route, perhaps we could use blank nodes to refer from each feature to other features defined elsewhere in the payload. But I am not sure this is a good idea.

More generally, if we start doing things the JSON-LD way, doesn't that vaguely imply that we should accept queries or responses which have a different JSON structure but are equivalent as far as their JSON-LD semantics are concerned? So forcing implementations to rely on some JSON-LD library to handle that somehow? I would be interested to know if there are specs which use JSON-LD and JSON schema simultaneously.

acka47:
> Ok, so I guess instead of `https://openrefine.org/context.jsonld`, services would use a URL of their own, which would describe their own features. But then, we are back to the problem of infinite feature sets. If your features can be parametrized (as they would be for Wikidata), then you cannot list them all in this context document. So you would need to automatically generate context documents which contain the metadata of the feature ids used in a particular candidate.

I apparently misunderstood what is modeled here. I thought we were talking about a controlled list of features that is the same for all reconciliation services. If the list differs from service to service, I have my doubts whether a JSON-LD context is the best solution.

> If we want to save space and go down the JSON-LD route, perhaps we could use blank nodes to refer from each feature to other features defined elsewhere in the payload. But I am not sure this is a good idea.

I don't understand what you mean.

> More generally, if we start doing things the JSON-LD way, doesn't that vaguely imply that we should accept queries or responses which have a different JSON structure but are equivalent as far as their JSON-LD semantics are concerned? So forcing implementations to rely on some JSON-LD library to handle that somehow? I would be interested to know if there are specs which use JSON-LD and JSON schema simultaneously.

I think it is no problem and even the most sensible solution to specify the API (by a JSON Schema or otherwise) in a way that only allows for one JSON-LD expression of the semantics. This would basically mean to define a JSON API with an added JSON-LD context that enables handling the data as RDF (or providing documentation for the properties and types). For example, [IIIF](https://iiif.io/) is doing it like that, see e.g. https://iiif.io/api/annex/notes/jsonld/#semantic-versioning. (A normative JSON schema for IIIF doesn't exist, though.)

wetneb:
> If the list differs from service to service, I have my doubts whether a JSON-LD context is the best solution.

Yes I would not want to specify the list of allowed features in the specs.

> I think it is no problem and even the most sensible solution to specify the API (by a JSON Schema or otherwise) in a way that only allows for one JSON-LD expression of the semantics. This would basically mean to define a JSON API with an added JSON-LD context that enables handling the data as RDF (or providing documentation for the properties and types). For example, IIIF is doing it like that, see e.g. https://iiif.io/api/annex/notes/jsonld/#semantic-versioning. (A normative JSON schema for IIIF doesn't exist, though.)

Ok, IIIF's stance on this makes sense. In that case I guess it would be safer not to rely on JSON-LD's technology to handle indirections: JSON-LD annotations could be added to the existing JSON payloads to document them, but these could also be safely ignored.

wetneb:
After many months I have added support for the `feature_view` suggested by @fsteeg. I prefer this over a preview-style service because I think it is not ideal to build a spec based on iframes in 2020… It's cleaner to just associate a URI to a feature, and let clients decide what they do with this URI.

wetneb:
Since the discussion has settled, I propose to merge this soon. We can have follow-up PRs if other issues arise.

PR #41, opened 2020-02-24
Authentication support with security schemes

wetneb:
This is another proposal for #26 where the specification of authentication is based on OpenAPI's notion of security schemes.
This lets services choose between api keys and HTTP auth. OpenID and OAuth are left out for now, they can be added to the specs later on if we are happy with the approach.

One downside of this is that there is no way to specify an optional authentication.


wetneb:
For the sake of making progress on this, I would propose introducing a separate boolean parameter (outside of the `"authentication"` field) to indicate an optional authentication.

ltrgoddard:
Hi all. In the case of authentication via an API key when using POST, should we consider whether it's more natural to pass the key as a parameter in the body of the request rather than as a query parameter (as would be usual with GET)? https://github.com/reconciliation-api/specs/pull/41/commits/620b65c8832c187b7f23d74ba7d66ebd10e53153 describes it as a "GET parameter".

OpenCorporates' implementation of the spec currently uses a query parameter for the API token regardless of the request method—I feel that having all parameters either in the URL (GET) or the body (POST) would be cleaner, but may be missing details about the implementation on the server side!

wetneb:
Good point. I guess, if we just rely on OpenAPI's definition of authentication methods, then we should just look up what they say on the matter.

I looked here but didn't really find something explicit:
https://swagger.io/docs/specification/authentication/api-keys/

That being said, we have noticed in the past that the way we POST query batches with a url-encoded JSON string is not very natural in OpenAPI: we should rather post the JSON only (with the right headers indicating the content type). If we do that, then it becomes much less natural to add the API key to the POST body (since it would be quite awkward to add it inside the JSON object itself, I think). So perhaps that tends to indicate that API keys should stay as GET parameters? Maybe it's a bit far-fetched.

ltrgoddard:
Thanks—I've just read some old issues and see now that this has come up before (https://github.com/reconciliation-api/specs/pull/37). I can't see anything explicit in the OpenAPI docs distinguishing between authentication for GET and POST either, but "query string" is a fairly clear term and if there's a chance of moving to JSON-only request payloads then I agree that it wouldn't make sense to use the request body for auth. I like the header method personally!

wetneb:

thadguidry (approved):
This looks good.  I was a bit confused initially with the simple `authentication` named object compared to [[OPENAPIS]] but now I understand and the examples match the [[OPENAPIS]] spec when I checked.  Really glad you linked a reference to it; that's what I used to eliminate my confusion, so other devs will find that useful I'm sure.

PR #48, opened 2020-07-21
Let services return descriptions in recon and data extension responses

### LINKED ISSUE #47: Add Description in Reconciliation Response
(This is a suggestion, based on my use of OpenRefine with Wikidata. I am not sure how well it generalises, nor am I aware of all your previous considerations. So feel free to just close this if I'm not making sense)

I am currently slowed down far more than necessary in my workflow by the inability to distinguish different suggestions returned by the Wikidata reconciliation service. This is especially prominent with location names, which seem to be less unique than personal names. Wikidata also has issues with duplicates, and specifically the common duplication (/triplication) of a location's name between the name of the place, its administration, and often the next-higher administrative level.
 
As it is, this requires either manually triggering a hover effect for each suggestion, or to implement a complex workflow of generating columns for all suggestions and then adding columns for each with the description fetched from the service. 

Wikidata (rightly) keeps disambiguation information out of labels and records it in the description. This requires the description to be passed along, and to be shown, wherever a user is asked to chose between alternatives. Not doing so is already [putting pressure on Wikidata to compromise its labels. ](https://www.wikidata.org/wiki/Wikidata:Project_chat/Archive/2020/07#Klammerzusatz_hinter_Vornamen_und_Familiennamen)

Therefore, I'd like to suggest adding the description as an optional field for the reconciliation batch response. This also brings it closer to the suggestion response, possibly making life easier for systems providing both.

wetneb:
Closes #47.

Descriptions were already supported by the suggest services, this lets services return them in reconciliation and data extension endpoints.

@MatthiasWinkelmann does this fit your bill?

### DIFF
--- a/latest/examples/data-extension-response/valid/example-full.json
+++ b/latest/examples/data-extension-response/valid/example-full.json
@@ -35,7 +35,8 @@
       "professionOrOccupation": [
         {
           "id": "4037223-6",
-          "name": "Malerin"
+          "name": "Malerin",
+          "description": "Beruf"
         },
         {
           "id": "4033430-2",
--- a/latest/examples/reconciliation-result-batch/valid/example-full.json
+++ b/latest/examples/reconciliation-result-batch/valid/example-full.json
@@ -4,6 +4,7 @@
       {
         "id": "120333937",
         "name": "Urbaniak, Regina",
+        "description": "1969-| Diss. Fachbereich Mathematik",
         "score": 53.015232,
         "match": false,
         "type": [
@@ -20,6 +21,7 @@
       {
         "id": "1127147390",
         "name": "Urbaniak, Jan",
+        "description": "Universität Wrocław, Niederlandestudien",
         "score": 52.357353,
         "match": false,
         "type": [
@@ -40,6 +42,7 @@
       {
         "id": "123064325",
         "name": "Schwanhold, Ernst",
+        "description": "1948-| Mitglied des Deutschen Bundestages, SPD (1993)",
         "score": 86.43497,
         "match": true,
         "type": [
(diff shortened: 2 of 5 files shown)


jetztgradnet:
Being "officially" able to return the description is great, thanks for that!
Regarding the suggest services, this requires another roundtrip and does not allow pro-actively returning additional data for reconciliation candidates.
For the purpose of this PR I am ok with that, but should we have another discussion/ticket for more open approach, e.g. with a sub-section in the candidate JSON structure which is not limited to a fixed set of pre-defined attributes?

wetneb:
Yes it is totally worth discussing making it possible to return user-configurable data fields. I am not sure I would add this to the suggest service, because these endpoints are generally designed to run very quickly and are only there to power auto-complete widgets, but it feels pretty suitable for the reconciliation queries.

MatthiasWinkelmann:
That.. was fast! Thanks a lot & I'll try to bring my own PR next time.

tfmorris:
Is the assumption that reconciliation services are monolingual? If not, the description should include a language tag and the client should have a way of requesting a language (or ordered list of languages), including the possibility of requesting None if they don't want the payload overhead.

wetneb:
Our names (labels) are monolingual at the moment so I would find it curious to have language tags for descriptions only, no? I agree it is a discussion worth having (it would be nice to have a single Wikidata recon endpoint, without having to insert a language code in it), but it seems orthogonal to this change. If services are to be language independent, I would introduce that for names and descriptions everywhere in the API, at the same time.

wetneb:
Merging this soon unless other comments are brought up in the meantime :)

PR #53, opened 2020-10-07
Make 'query' field in reconciliation queries optional.

### LINKED ISSUE #27: Reconciling without names
The reconciliation API assumes that reconciliation queries contain at
least a name (as basic search query). It is then possible to add other
constraints (type, properties). However in many situations the user does
not have any name to supply.

The most common situation where this is a problem is when the user has
access to a unique identifier, supported by the reconciliation service
as a property. Supplying such a unique identifier as a property in a
reconciliation query should be enough to identify the candidates, but in
the current API a name also has to be provided.

We should make it easier to handle such situations in the API.

Mailing list thread: https://lists.w3.org/Archives/Public/public-reconciliation/2019Nov/0002.html

wetneb:
Fixes #27.

This just makes it possible to send queries such as `{"properties":[{"pid":"my_id", "v":"394adf1ffb03"}]}`, where no `"query"` for the name is supplied. It seems fairly straightforward as far as the specs are concerned (that does not mean I have a clear picture of how this should be exposed in OpenRefine).

Another way to do this would be to move the existing `"query"` field into the `"properties"` array, by giving it a dedicated `"pid"` in the specs, but that seems less backwards-compatible, more verbose and offering no particular benefit.


latest/index.html
```diff
@@ -372,26 +372,27 @@ <h2>Reconciliation Queries</h2>
       <section>
         <h3>Structure of a Reconciliation Query</h3>
         <p>
-          A <dfn>reconciliation query</dfn> consists of:
+          A <dfn>reconciliation query</dfn> consists of the following fields. They are all optional
+	  but at last one of <code>query</code> and <code>properties</code> must be supplied.
```
tfmorris, line 376:
```suggestion
          A <dfn>reconciliation query</dfn> consists of the following fields. 
	  At least one of <code>query</code> or <code>properties</code> must be supplied, but all other
	  fields are optional.
```

latest/index.html
```diff
@@ -372,26 +372,27 @@ <h2>Reconciliation Queries</h2>
...
+            <dd>A query string, consisting of a non-empty string.
 	    	By supplying such a string, a client intends to search for entities with similar
 	    	names. The specifics of how this similarity is defined are determined by the service.</dd>
 	    <dt><code>type</code></dt>
-            <dd>Optionally, a list of <a>types</a>. Supplying such types allows users to restrict
+            <dd>A list of <a>types</a>. Supplying such types allows users to restrict
 	    	the search to entities which bear those types. Whether this restriction should be a
 	    	hard constraint or simply induce a change on the reconciliation scores can be
 	    	determined by the service. In particular, services MAY return candidates which
 	    	do not belong to any of the supplied types;</dd>
```
tfmorris, line 387: Unrelated to this PR, but services may only do this if allowed by `type_strict == "should"`, otherwise they need to consider the `all` or `any` type criteria.

tfmorris (changes requested):

tfmorris (approved):
LGTM

wetneb:
Merging since this has been open for a long time with no outstanding concerns :)

PR #65, opened 2021-02-07
Make manifest access more explicit.

### LINKED ISSUE #64: When to supply the service manifest?
From [this example](https://github.com/mikejs/reconcile-demo/blob/master/reconcile.py) I understand that all interaction comes from a single endpoint and that the service manifest is to be delivered from the end point if it's not a query.  I think it might be worth making that explicit in the API.  If this is reason for leaving this ambiguous, then I think an explanation of why would be helpful.  Should the service manifest always be accessible with a GET request?

wetneb:
Closes #64.


tfmorris (approved):
LGTM

Issue #66, opened 2021-02-20
Suggest query batch sizes and HTTP concurrency


wetneb:
Reconciliation clients can submit reconciliation query batches but there is no clear indication of how big these batches should be. Also, there is no clear indication of how many parallel HTTP requests should be fired at a given service.

The specs could be more precise on this, perhaps by letting services announce in their manifests some maximum or recommended values for these settings.

Corresponding OpenRefine issues: https://github.com/OpenRefine/OpenRefine/issues/2549, https://github.com/OpenRefine/OpenRefine/issues/3657.

tfmorris:
"Concurrency" is just, in my opinion, a special case of rate limiting, which is already covered by #40 (and #43).

I'm not sure anything that is service level related should go in the manifest, since it may vary by the service class of the user, be load related, or by specified out of band.

If there's a maximum batch size, I guess I could see adding that to the manifest (although it's also affected by a variety of HTTP parameters), but would it affect what clients' behavior? They are more likely to choose a batch size which is in the sweet spot that matches their desires for latency, reliability, etc which is probably much smaller than any reasonable maximum.

More generally, perhaps we should try to standardize some core functionality, get some experience with it, and then see what's high priority to add.

gitonthescene:
Concurrency is about parallelism and rate limiting is not.  I agree that the service doesn't have much control of what a client may attempt and that rate limiting is one way to handle a poorly acting client.  On the other hand, the service manifest is one method that the server can declare what it believes is acceptable from a client.  Different services may have different levels of tolerance for load and so the client doesn't need to treat all servers alike.  Currently, OpenRefine sends reconciliation requests with a maximum batchSize of 10 regardless of the server.  Similarly, OpenRefine sends only one request at a time regardless of how long it takes to process a request.  Concurrency allows services to work on multiple requests in parallel.  The server doesn't necessarily know that the requests came from the same client without sessions, but can rate limit based on IP address.  In general, I don't think the service should expect all clients to be good actors and should be configured defensively and similarly the client shouldn't make assumptions about what it takes to be a good actor.  The client should take hints from the service via the service manifest and react to errors with a back-off on retries.

gitonthescene:
[...]

PR #67, opened 2021-02-22
Add info on rate limit contracts for batches

gitonthescene:
Updated spec as requested in https://github.com/OpenRefine/OpenRefine/issues/3657.  Referred to in #66.

### DIFF
--- a/latest/index.html
+++ b/latest/index.html
@@ -337,6 +337,10 @@ database are instances of this type.<dd>
           </dd>
           <dt><code>extend</code></dt>
           <dd>A <a>data extension metadata</a>, supplied if the service offers a <a href="#data-extension-service">data extension service</a>.</dd>
+          <dt><code>batchSize</code></dt>
+          <dd>The maximum number of <a>reconciliation queries</a> in a single <a>reconciliation query batch</a>.  The service MAY respond to batches larger than this number with a 413 HTTP error status code [[RFC7231]]</dd>
+          <dt><code>batchRate</code></dt>
+          <dd>The number of <a>reconciliation query batch</a> requests which can safely be sent to this service per second per IP address.  The service MAY respond to requests above this rate with a 503 HTTP error status code [[RFC7231]].</dd>
         </dl>
       </p>
       <p>For instance, a service could expose the following minimal service manifest:


wetneb:
Thanks! I think this goes in the right direction :)
Here are a few thoughts:
* perhaps these parameters could be grouped into a subobject, to keep the root of the manifest a bit cleaner?
* we would need to add these parameters to some JSON examples and to the JSON schemas for validation (I am not asking that you do this yet, since we should first have an agreement on the desired syntax).

gitonthescene:
Hey there @wetneb, I’m just getting back to this.  I’m not sure I understand what you’re thinking of in terms of a subsection.  Maybe you could offer a concrete suggestion.  Currently, it’s two new parameters that aren’t even necessarily related.

wetneb:
I was thinking that instead of specifying something like
```json
{
    "name": "my recon service",
    "identifierSpace": "my identifier space",
    "schemaSpace": "my schema space",
    "batchSize": 100,
    "batchRate": 10
}
```
we could instead have something like
```json
{
    "name": "my recon service",
    "identifierSpace": "my identifier space",
    "schemaSpace": "my schema space",
    "limits": {
        "batchSize": 100,
        "batchRate": 10
    }
}
```
For me, those parameters are topically related and we might want to introduce more of them in the future, so it feels natural to try to group them together :)

gitonthescene:
I don't mind doing it that way, though I suspect "limits" may be a bit too broad a category to add any cognitive value.  I think rate limiting is likely to come with more parameters eventually that have nothing to do with the batch size.

gitonthescene:
Separately, thinking about how the spec might evolve raises the question about what's the intended process for servers to support a given version of a spec.  I.e. how is the version of the spec being used negotiated between the client and the server.  I see that the server can supply the list of versions it supports in the [service manifest](https://reconciliation-api.github.io/specs/latest/#service-manifest) but it's not clear how the handshake works.

For instance, if we introduced "batchRate" and later changed it to "throttleRate", how would the service manifest indicate which it was offering.  Or if it's able to offer one or the other according to what the client understands, how would it receive the information about what the client is expecting?

For this specific change the server could offer both fields (though that likely means it's not strictly following either spec), but if the meaning of a field changes then even that is not possible.

wetneb:
About the format, my example above was just meant to illustrate what I meant by "subobject": if you can think of a better term to group these two settings, I would be glad to hear it.

About versioning, I suggest we have this discussion somewhere else :)

gitonthescene:
@wetneb I don't know that there is something which applies at this point.  I'm happy with my original suggestion.  Should we just go with that unless you can think up something better?

wetneb:
I think I would still prefer to keep these options grouped, mostly for the sake of not having a plethora of root-level parameters in this JSON object. I think `limits` or `throughput` would both work, but I would be happy with other terms.

Beyond those petty syntax choices I think it would be worth thinking more about the parameters that are actually useful to the services here. In https://github.com/OpenRefine/OpenRefine/issues/3657 you mention the idea of sending all requests in parallel at once and I think that is precisely what many services would really not enjoy. In my experience many services ask clients to cap the number of parallel connections to the service (because open connections are a precious resource for servers).
So instead of / in addition to your `batchRate` I would be tempted to have something like `maxConnections` which would ask a client to cap the number of open sockets to the service.

gitonthescene:
I’m okay with a max connections parameter.  As I said in that issue the server should mostly be responsible for a defensive implementation and shouldn’t rely on clients playing nice.  Also parallelizing a request should allow it to complete quicker and thereby release resources.

But again, I’m mostly interested in batchSize to be able to horizontally scale the requests on the server side.  If the client only sends the current default of ten items at a time, not much can be done by the server to improve the throughput.

wetneb:
Other suggestion of names for this section of the manifest: `serverLimits`.

gitonthescene:
Sounds good to me.  Should we just go with that?  We can start with just batchSize and add to it as we go along.

wetneb:
I have built on this PR in #80 with the proposed consensus summarized by @fsteeg, so closing this one.

Issue #39, opened 2020-02-18
Placeholder for ids is not consistent between services


metaschell:
The placeholders used to embed an entity id in the service URLs defined in the manifest differ between the services. These should be consistent, ideally choosing `{{id}}` over `${id}` to avoid pitfalls in clients/language interpreting `$` as replacement character within strings (which is kind of the point here as well, but still using this without explicitly providing the character, e.g. in a bash with CURL might lead to subtle and hard to spot errors...)

The preview service (section 5.1) uses `{{id}}` as placeholder in the service description url:
https://reconciliation-api.github.io/specs/latest/#preview-metadata

The flyout service (sections 6.1 and 6.5) uses `${id}` as placeholder in the service description url:
https://reconciliation-api.github.io/specs/latest/#suggest-metadata
https://reconciliation-api.github.io/specs/latest/#flyout-services

This also affects the Manifest JSON schema defined in section A.1
https://reconciliation-api.github.io/specs/latest/#manifest-schema

wetneb:
That's also something I have found annoying. It would be great to unify this.

Also, we should make it clearer how URL escaping works in these patterns. If an id contains special characters, are they going to be escaped before being inserted in the template?

metaschell:
Escaping should be defined, especially when returning IRIs/URIs as IDs.

We are looking to implement the Reconciliation API with linked data, so the ID is always an IRI. Or rather the local name of an IRI as I understand it, as the namespace is declared in the service manifest? Which is a bit of a problem when searching through data from mixed sources which might have mixed/different namespaces...

wetneb:
If you are using entire IRIs as IDs, then you should be able to use templates without any prefix or suffix - that relies on the assumption that ids are not escaped when inserted in templates.

The `identifierSpace` defined in the manifest does not have any concrete influence on how ids are understood. You can just use any URI that describes your "identifier space" - it does not need to be a namespace in the sense of RDF.

On top of my head I don't remember exactly what OpenRefine currently does concerning escaping. (It's probably not worth following - we just need to think about what the specs should be in an ideal world and then update OpenRefine to follow that.)

metaschell:
That sounds like a good approach :-)

wetneb:
In OpenRefine we are considering introducing escaping for the ids, so that would potentially break @metaschell's use case of using full URIs as ids:
https://github.com/OpenRefine/OpenRefine/issues/3730
[...]

PR #71, opened 2021-06-15
URI component encoding for view, preview & flyout templates

fsteeg:
This corresponds to the implementation in https://github.com/OpenRefine/OpenRefine/pull/3731, which implements this for the view and preview (but not yet for the flyout) URL templates. See https://github.com/reconciliation-api/specs/issues/39#issuecomment-825570330 for the reasoning behind this approach (vs. supporting full URIs as identifiers).

### DIFF
--- a/latest/index.html
+++ b/latest/index.html
@@ -216,7 +216,8 @@ href="https://github.com/OpenRefine/OpenRefine/wiki/Reconciliable-Data-Sources">
            which associates to each entity a corresponding URI, by inserting its identifier in the template.
            A view template is a string which contains the <code>{{id}<!-- -->}</code> substring.
            For each entity, replacing <code>{{id}<!-- -->}</code> in the template by the entity's identifier
-           MUST result in a valid URI [[RFC2396]].
+           MUST result in a valid URI. To guarantee the correctness of the formed URI, clients MUST percent-encode
+           component-delimiting reserved characters in the identifier, i.e. encode it as a URI component [[RFC3986]].
         </p>
         <p>
            Similarly, it is possible to associate to each <a>matching feature</a> a URL where documentation about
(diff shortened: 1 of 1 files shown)


latest/index.html
```diff
@@ -558,7 +559,7 @@ <h3>Preview Metadata</h3>
       <p>Reconciliation services MAY offer a preview service by providing the <dfn>preview metadata</dfn> as an object stored in the <a>service manifest</a> under the key <code>preview</code>. It consists of the following fields, all mandatory:
         <dl>
           <dt><code>url</code></dt>
-          <dd>A string containing <code>{{id}<!-- -->}</code> such that replacing <code>{{id}<!-- -->}</code> by an <a>entity</a> identifier yields the <dfn>preview URL</dfn> for that entity. This preview URL MUST resolve to an HTML page summarizing the entity.
+          <dd>A string containing <code>{{id}<!-- -->}</code> such that replacing <code>{{id}<!-- -->}</code> by an <a>entity</a> identifier encoded as a URI component yields the <dfn>preview URL</dfn> for that entity. This preview URL MUST resolve to an HTML page summarizing the entity.
```
thadguidry, line 562: Should we say "percent-encoded" or just "encoded" here?

wetneb: I think "encoded as a URI component" is quite precise, no?

wetneb (approved):

latest/index.html
fsteeg: My idea was to have the details in the first mention in the entity section (which contains "percent-encode"), basically defining what "encode as a URI component" means, and then using that short form.

thadguidry: @fsteeg OK sure no problem. RFC3986 continually used the long form.

wetneb:
Merging this since Thad's comment was addressed.

Issue #68, opened 2021-04-13
Representing type hierarchies


wetneb:
In some reconciliation services (such as Wikidata), the types are organized in a hierarchy: some types are included in others. For now, the specifications do not say anything about this. In today's call, it was mentioned that the specifications could also say something about that. What would be the impact on the API? What workflows would it enable?

fsteeg:
After our [last meeting](https://etherpad.lobid.org/p/reconciliation-june-2021), this is what I understand about the use case here. We can currently choose a suggested default type in the reconciliation dialog (1) or search and get suggestions for a specific type (2):

![recon-types](https://user-images.githubusercontent.com/128229/125959603-21936acc-d919-4ff5-aee0-5e9b6a3faef1.png)

So the idea here would be to additionally provide a way visually browse a type hierarchy for selecting a type to reconcile against. This would be very nice to have for [the GND class/type hierarchy](https://blog.lobid.org/2018/03/20/gnd-class-hierarchy.html) too. It would provide a way to discover what types are available in the dataset, where today we can only discover default types that are suggested (1) or that we already know of and search for specifically (2).

On the protocol level, this would require expressing the hierarchy somehow when describing [types](https://reconciliation-api.github.io/specs/latest/#types). This could be achieved by adding fields like `broader` and/or `narrower` to each type (like [in SKOS](https://www.w3.org/TR/skos-reference/#semantic-relations)). These fields could be ignored by current clients/functionality (like (1), using the `defaultTypes` in the [service manifest](https://reconciliation-api.github.io/specs/latest/#service-manifest) and (2), using `type` [suggest responses](https://reconciliation-api.github.io/specs/latest/#suggest-responses)). New clients/functionality could use `broader` and/or `narrower` to build a visual representation of the hierarchy. To get the full hierarchy, they could use a `type` [suggest query](https://reconciliation-api.github.io/specs/latest/#suggest-queries) with an empty `prefix` param (or omit the `prefix` param).

wetneb:
My initial reaction to this proposal was that to make this work in Wikidata's case, we would need to add quite some paging mechanisms because Wikidata's type hierarchy is huge:
* the number of items that can be used as types is too large to be sent in one HTTP response
* the number of children (subclasses) of a given type is too large to be stored in full in one JSON object
[...]

PR #73, opened 2021-08-17
Representing type hierarchies

fsteeg:
Current work in progress for #68.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -9,8 +9,8 @@ for version in "0.1" "latest"; do
 	for schema_file in `ls $version/schemas`; do
 		echo "Testing $schema_file"
 		schema_name=$(echo "$schema_file" | cut -f 1 -d '.')
-		ajv test -s $version/schemas/$schema_file -d "$version/examples/$schema_name/valid/*.json" --valid
-		ajv test -s $version/schemas/$schema_file -d "$version/examples/$schema_name/invalid/*.json" --invalid
+		ajv test -s $version/schemas/$schema_file -r latest/schemas/type.json -d "$version/examples/$schema_name/valid/*.json" --valid
+		ajv test -s $version/schemas/$schema_file -r latest/schemas/type.json -d "$version/examples/$schema_name/invalid/*.json" --invalid
 		echo ""
 	done
 done
--- a/latest/examples/manifest/valid/lobid-gnd.json
+++ b/latest/examples/manifest/valid/lobid-gnd.json
@@ -31,7 +31,7 @@
     {
       "id": "DifferentiatedPerson",
       "name": "Individualisierte Person",
-      "broader": "Person"
+      "broader": ["Person"]
     },
     {
       "id": "Family",
--- a/latest/examples/manifest/valid/lobid-gnd.json
+++ b/latest/examples/manifest/valid/lobid-gnd.json
@@ -30,7 +30,8 @@
     },
     {
       "id": "DifferentiatedPerson",
-      "name": "Individualisierte Person"
+      "name": "Individualisierte Person",
+      "broader": "Person"
     },
     {
       "id": "Family",
--- a/latest/examples/manifest/valid/lobid-gnd.json
+++ b/latest/examples/manifest/valid/lobid-gnd.json
@@ -31,7 +31,12 @@
     {
       "id": "DifferentiatedPerson",
       "name": "Individualisierte Person",
-      "broader": ["Person"]
+      "broader": [
+        {
+          "id": "Person",
+          "name": "Person"
+        }
+      ]
     },
     {
       "id": "Family",
--- a/latest/examples/reconciliation-candidate/valid/example.json
+++ b/latest/examples/reconciliation-candidate/valid/example.json
@@ -28,7 +28,13 @@
     },
     {
       "id": "DifferentiatedPerson",
-      "name": "Individualisierte Person"
+      "name": "Individualisierte Person",
+      "broader": [
+        {
+          "id": "Person",
+          "name": "Person"
+        }
+      ]
     }
   ]
 }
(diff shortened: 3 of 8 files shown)


fsteeg:
As a basic test case for this, I've added single-level `broader` type support to our reconciliation service at https://test.lobid.org/gnd/reconcile (see https://github.com/hbz/lobid-gnd/commit/5c8d0a8a2049b06dee23827e6e33cf20065b3f1f) and to the testbench (see https://github.com/reconciliation-api/testbench/pull/28). I think this shows how even a simple implementation of `broader` IDs can be useful. In the future, we could look into full hierarchy support in the testbench, if that makes sense.

wetneb (approved):
Amazing! Nice that you managed to update the testbench and your endpoint in the same go. It looks great to me, but maybe we can wait for our call later today to merge this, in case others have more feedback.

orothad:
```json
{
  "$schema": "http://json-schema.org/schema#",
  "$id": "https://reconciliation-api.github.io/specs/latest/schemas/suggest-types-response.json",
  "type": "object",
  "description": "This schema can be used to validate the JSON response of a suggest service for types.",
  "properties": {
    "result": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string",
            "description": "Identifier of the suggested type"
          },
          "name": {
            "type": "string",
            "description": "Name of the suggested type"
          },
          "description": {
            "type": "string",
            "description": "An optional description which can be provided to disambiguate namesakes, providing more context."
          },
          "broader": {
            "type": "string",
            "description": "A broader semantic relation as defined by SKOS http://www.w3.org/2004/02/skos/core#broader",
            "$id": "http://www.w3.org/2004/02/skos/core#broader",
            "additionalProperties": {
              "label": "skos:broadMatch"
            }
          }
        },
```

Some clients might want to enrich the presentation of properties with additional metadata or annotations.
In the above example "broader" is given more context and where that context can be displayed by a smart client.

I read though some of the JSON Schema core and appendix for Working with vocabularies, but not 100% sure my example is conforming for validation, but still wanted to show what I was asking about on our last meeting, that of giving more context for a property such as "broader".  And it seems JSON Schema has various ways that more context, annotation, or metadata could be given depending on validation needs, or ignoring validation (the validator ignores, but smart clients can still understand and use)

1. I "think" in the above case that "label" would be ignored by some clients? - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2.3

2. I'm not sure about "$id" but it seems to be valid for JSON Schema? - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.2.1

thadguidry:
```json
{
  "$schema": "http://json-schema.org/schema#",
  "$id": "https://reconciliation-api.github.io/specs/latest/schemas/suggest-types-response.json",
  "type": "object",
  "description": "This schema can be used to validate the JSON response of a suggest service for types.",
  "properties": {
    "result": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string",
            "description": "Identifier of the suggested type"
          },
          "name": {
            "type": "string",
            "description": "Name of the suggested type"
          },
          "description": {
            "type": "string",
            "description": "An optional description which can be provided to disambiguate namesakes, providing more context."
          },
          "broader": {
            "type": "string",
            "description": "A broader semantic relation as defined by SKOS http://www.w3.org/2004/02/skos/core#broader",
            "$id": "http://www.w3.org/2004/02/skos/core#broader",
            "additionalProperties": {
              "label": "skos:broadMatch"
            }
          }
        },
```

Some clients might want to enrich the presentation of properties with additional metadata or annotations.
In the above example "broader" is given more context and where that context can be displayed by a smart client.

I read though some of the JSON Schema core and appendix for Working with vocabularies, but not 100% sure my example is conforming for validation, but still wanted to show what I was asking about on our last meeting, that of giving more context for a property such as "broader".  And it seems JSON Schema has various ways that more context, annotation, or metadata could be given depending on validation needs, or ignoring validation (the validator ignores, but smart clients can still understand and use)

1. I "think" in the above case that "label" would be ignored by some clients? - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2.3

2. I'm not sure about "$id" but it seems to be valid for JSON Schema? - https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.8.2.1

fsteeg:
Switched from array of strings to array of types in https://github.com/reconciliation-api/specs/pull/73/commits/f19ce913afe4e9efadf06e42ada07ca321ceac20 as discussed in https://github.com/reconciliation-api/testbench/pull/28 and https://etherpad.lobid.org/p/reconciliation-november-2021.

Next: update our service implementation in https://github.com/hbz/lobid-gnd/pull/293 and testbench in https://github.com/reconciliation-api/testbench/pull/28.

fsteeg:
> Next: update our service implementation [...] and testbench [...].

Done, see https://github.com/hbz/lobid-gnd/pull/293#issuecomment-993368743 and https://github.com/reconciliation-api/testbench/pull/28#issuecomment-993370187.

test.sh
```diff
@@ -9,8 +9,8 @@ for version in "0.1" "latest"; do
 	for schema_file in `ls $version/schemas`; do
 		echo "Testing $schema_file"
 		schema_name=$(echo "$schema_file" | cut -f 1 -d '.')
-		ajv test -s $version/schemas/$schema_file -d "$version/examples/$schema_name/valid/*.json" --valid
-		ajv test -s $version/schemas/$schema_file -d "$version/examples/$schema_name/invalid/*.json" --invalid
+		ajv test -s $version/schemas/$schema_file -r latest/schemas/type.json -d "$version/examples/$schema_name/valid/*.json" --valid
+		ajv test -s $version/schemas/$schema_file -r latest/schemas/type.json -d "$version/examples/$schema_name/invalid/*.json" --invalid
```
wetneb, line 13: I am wondering if there would be a more flexible way to do this… I assume you tried and we can just go with this for now, and think of a better way once this becomes a problem.

wetneb (approved):

PR #80, opened 2021-12-14
Adds ability to announce some rate-limiting settings

wetneb:
Builds on #67 and adds the proposed syntax to the JSON schema and examples.


wetneb:
I wonder if it would make sense to move the `batchSize` attribute inside this new subobject. I would assume it is fine to do this because the `batchSize` parameter has not been included in any official version of the specs yet.

wetneb:
I have tried to align this PR with the consensus described by @fsteeg at https://github.com/reconciliation-api/specs/issues/66#issuecomment-1161603702.

wetneb:

PR #85, opened 2022-05-30
Switch to RFC6570 for URI templates

### LINKED ISSUE #39: Placeholder for ids is not consistent between services

wetneb:
Closes #39.

This switches to the RFC6570 syntax for URI templates, as suggested in #39 by @Robsteranium.
This has the benefit of not forcing URL encoding of the identifiers when inserting them in the template, and allowing for a much wider range of use cases. This is a significant change since the syntax used so far (`{{id}}` in most places, `${id}` for the flyout template) are not compatible with RFC6570. However I think it is really important to support the use case of services using a full URI as identifier (since that is already happening in the wild), or parts of URI which should be inserted without escaping.

### DIFF
--- a/latest/examples/manifest/valid/codefork-viaf.json
+++ b/latest/examples/manifest/valid/codefork-viaf.json
@@ -4,7 +4,7 @@
   "identifierSpace": "http://rdf.freebase.com/ns/user/hangy/viaf",
   "schemaSpace": "http://rdf.freebase.com/ns/type.object.id",
   "view": {
-    "url": "http://viaf.org/viaf/{{id}}"
+    "url": "http://viaf.org/viaf/{id}"
   },
   "defaultTypes": [
     {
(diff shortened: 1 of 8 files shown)


fsteeg:
Makes sense and looks good to me, +1.

wetneb:
Since this is not backwards compatible, in our last call we suggested to wait for a 0.2 release before merging this.

wetneb:
Merging now that 0.2 is released.

PR #87, opened 2022-06-03
Add support for returning entities that are understood by other services

### LINKED ISSUE #75: Federation between services via data extension
In the OpenRefine project we are currently developing a reconciliation service for [Wikimedia Commons](https://commons.wikimedia.org/).

Wikimedia Commons uses [Wikidata](https://www.wikidata.org/) items and properties to describe the media files it stores. Therefore, our data extension service on the Commons endpoint will be able to return Wikidata items (as values of properties on Commons entities). It would be nice if our Commons service could somehow announce that these returned entities are in Wikidata, for instance by declaring that they are in the entity space of the Wikidata reconciliation service.

This would theoretically allow for workflows where users could first fetch data from the Commons service, and then fetch more data from the Wikidata recon service (with the data extension mechanism again).

We could use the `meta` part of data extension responses to declare a reconciliation endpoint and/or identifierSpace/schemaSpace.

wetneb:
This adds a new field to data extension responses, to let them return entities which belong to the identifier space of another service. I made the choice of returning a reconciliation endpoint rather than an identifier space, because I think this is going to be more useful to users (if they want to keep doing data extension with the new endpoint). But I think one can also make a case for supplying the identifier space (to let the user pick which reconciliation service they want to use for this identifier space).

Closes #75.

### DIFF
--- a/latest/examples/data-extension-query/valid/example-full.json
+++ b/latest/examples/data-extension-query/valid/example-full.json
@@ -19,6 +19,9 @@
     },
     {
       "id": "professionOrOccupation"
+    },
+    {
+      "id": "wikidataId"
     }
   ]
 }
--- a/latest/examples/data-extension-response/valid/example-full.json
+++ b/latest/examples/data-extension-response/valid/example-full.json
@@ -15,6 +15,11 @@
         "id": "SubjectHeading",
         "name": "Schlagwort"
       }
+    },
+    {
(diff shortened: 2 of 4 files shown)


wetneb:
That is the intention indeed. For instance, being able to use the preview or data extension functionalities of the referred service.

wetneb:

PR #89, opened 2022-07-12
Add 'documentation', 'serviceVersion' and 'logo' fields in manifest.

### LINKED ISSUE #86: More metadata fields in manifests: icon, documentation
In an API client like OpenRefine, it would be nice if we could:
* have a logo for each reconciliation service that has been added by the user. This suggests a standard way to obtain a logo for a service, either by supplying it as a field in the manifest, or by serving it at a standard location relative to the manifest (such as a favicon)
* offer a link to the documentation of the service, for instance to find out what properties or types it supports;
* anything else?

Intuitively it does not hurt to add optional support for such things: that does not make the life of service developers harder, and can yield nice benefits for the user if those mechanisms end up being adopted widely.

wetneb:
Closes #86.

### DIFF
--- a/latest/index.html
+++ b/latest/index.html
@@ -321,6 +321,12 @@ will help disambiguating it from others;</dd>
           <dt><code>defaultTypes</code></dt>
           <dd>An array of <a>types</a> which are considered sensible default choices as types supplied in reconciliation queries. For services which do not rely on types, this MAY contain a single type with a generic name making it clear that all entities in the
 database are instances of this type.<dd>
+          <dt><code>documentation</code></dt>
+          <dd>An optional URL with human-readable documentation about the service, for instance giving more information about the data it exposes;</dd>
+          <dt><code>logo</code></dt>
+          <dd>An optional URL of a square image which can be used as the service's logo;</dd>
+          <dt><code>serviceVersion</code></dt>
+          <dt>An optional string which describes the version of the software exposing this service. This is not to be confused with <code>versions</code> which is about the versions of the protocol supported by the service;</dt>
           <dt><code>view</code></dt>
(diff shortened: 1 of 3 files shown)


wetneb:

latest/index.html
```diff
@@ -321,6 +321,12 @@ <h3>Service Manifest</h3>
           <dt><code>defaultTypes</code></dt>
           <dd>An array of <a>types</a> which are considered sensible default choices as types supplied in reconciliation queries. For services which do not rely on types, this MAY contain a single type with a generic name making it clear that all entities in the
 database are instances of this type.<dd>
+          <dt><code>documentation</code></dt>
+          <dd>An optional URL with human-readable documentation about the service, for instance giving more information about the data it exposes;</dd>
+          <dt><code>logo</code></dt>
+          <dd>An optional URL of a square image which can be used as the service's logo;</dd>
+          <dt><code>serviceVersion</code></dt>
+          <dt>An optional string which describes the version of the software exposing this service. This is not to be confused with <code>versions</code> which is about the versions of the protocol supported by the service;</dt>
```
thadguidry, line 329:
```suggestion
          <dt>An optional string which describes the version of the software exposing this service. This is not to be confused with <code>versions</code> which is about the versions of the reconciliation api supported by the service;</dt>
```

thadguidry: I think saying `protocol` felt slightly out of place here.  If I'm wrong, that's fine.

thadguidry (changes requested):

latest/index.html
wetneb: Fine with me!

Issue #84, opened 2022-05-19
Make the API more REST-y


wetneb:
I just had a very nice conversation with @stefanw who told me about his experience of implementing the reconciliation API for FragDenStaat.de. They use the Django REST Framework and he tried to implement the protocol in this context. You can see the source code here: https://github.com/okfde/froide/blob/main/froide/helper/api_utils.py#L104

One general issue that he has encountered is that the API does not really follow the REST principles. We depart from those principles in the following ways (on top of my head), with the checkbox indicating if the problem has been fixed already:
* [x] the use of JSONP (#19)
* [x] the use of verbs (if your endpoint is at `/api`, then there is a big difference in what `/api`,  `/api?queries=` and `/api?extend=` do) (#17)
* [x] the use of URL-encoded JSON payloads (POSTing `?queries=%7B%22q0%22%3A%7B%22query%22%3A%22foo%22%7D%7D` instead of just `{"q0":{"query":"foo"}}` with `Content-Type: application/json`) (#32)
* [x] the use of varying endpoints defined in the manifest (#17)
* [x] #33

and perhaps other things I have missed?

This is not the first time I have heard this, so I think it would make sense to fix those problems, even though they induce breaking changes. Versioning should make it possible for clients such as OpenRefine to keep supporting the old API and the new one.

wetneb:
Yes, I think it would be important that OpenRefine gets updated quickly to support the new version too. I also do not expect many services to adopt the new specs until that is done, since it remains the main client for this API as far as I can see.

Although I cannot commit to a specific time-frame yet, I would generally be keen to work on this myself, or helping others to do it. In any case, I would definitely expect that there is a period during which the REST-ified API has been already published but the latest stable version of OpenRefine does not support it yet - I would like to make this period short, but I think it is also fine if it lasts a few months. Or do you think this is not acceptable?

stefanw:
Since I got tagged, here are my two cents as an implementer of the API:

- I would have gladly used any kind of library that could have made the implementation easier. Something where I just need to implement some methods and it does the rest for me. Writing the parsing for the API requests was cumbersome, had lots of difficult to debug errors and took a long time. (I hear it's gotten better since then with more tooling.)
- Reconciliation seems slow to me and I think it's because it's serialized and non-batched. Reconciling batches of records at once could lower the overhead per HTTP request. Most backends are probably fast enough to handle reconciliation of hundreds of records in one HTTP request. This would probably not be very RESTy, but maybe that's OK.
[...]

PR #91, opened 2022-07-26
Impose pre-determined routes instead of announcing them in the manifest

### LINKED ISSUE #32: Submitting queries as JSON without URL-encoding them

### LINKED ISSUE #42: Remove flyout services
Now that the suggest services return descriptions (which was not the case a few years ago), we could consider removing the flyout service which is redundant with these descriptions. Flyout services lead to an awkward user experience where the user must hover a suggest drop-down menu to display the description:
image

See related issue in Wikidata recon service: https://github.com/wetneb/openrefine-wikibase/issues/77

wetneb:
This is a first step towards #84: instead of leaving the suggest, preview and property proposal paths up to the implementer, those paths are pre-determined by the spec.

This was also the occasion to remove the flyout: I had to do this when changing the manifest structure. Sorry to be bundling that up in the same PR, I should have done another one about this earlier. Therefore it closes #42.

It also closes #32 in the same go.

### DIFF
--- a/latest/index.html
+++ b/latest/index.html
@@ -584,7 +584,7 @@ in the <code>score</code> field). By exposing individual features in their respo
 	<p>
           A reconciliation service MUST support HTTP POST requests at the route <code>/reconcile</code> (relative to its <a>endpoint</a>) with
           <code>application/json</code> bodies containing a
-          <a>reconciliation query batch</a> (serialized in JSON) in a form element named <code>queries</code>.
+          <a>reconciliation query batch</a>.
         </p>
         <p>
           <pre class="example nohighlight">POST /reconcile &lt;reconciliation query batch&gt;</pre>
--- a/latest/index.html
+++ b/latest/index.html
@@ -639,9 +639,9 @@ in the <code>score</code> field). By exposing individual features in their respo
       <p>Reconciliation services MAY offer a preview service by providing the <dfn>preview metadata</dfn> as an object stored in the <a>service manifest</a> under the key <code>preview</code>. It consists of the following fields, all mandatory:
         <dl>
           <dt><code>width</code></dt>
-          <dd>The width in pixels of the <code>&lt;iframe&gt;</code> element where to render an entity preview;</dd>
+          <dd>The width in pixels of the viewport where to render an entity preview;</dd>
           <dt><code>height</code></dt>
-          <dd>The height in pixels of the same <code>&lt;iframe&gt;</code>.</dd>
+          <dd>The height in pixels of the same viewport.</dd>
         </dl>
       </p>
       <p>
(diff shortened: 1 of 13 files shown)


latest/index.html
```diff
@@ -638,10 +651,11 @@ <h3>Preview Metadata</h3>
       </section>
       <section>
       <h3>Preview Queries</h3>
-      <p>A preview service is queried by resolving the <a>preview URL</a> for an <a>entity</a>. The URL must resolve to an HTML document.
+      <p>A preview service is queried by resolving the URI template <code>/preview?id={id}</code> relative to the reconciliation endpoint, where <code>id</code> is subsituted by the <a>entity</a> identifier. The URL must resolve to an HTML document,
+        which MUST be viewable in an <code>iframe</code> HTML element whose dimensions are determined by the <a>preview metadata</a>.
```
thadguidry, line 655: Shouldn't that be CAN instead?  I mean some clients don't necessarily need an iframe, it's up to them how to interact with data coming from a standardized api spec?

wetneb: That `MUST` does not force clients to embed the preview in an iframe, it only ensures that the server makes that possible. I would be fine with reformulating this to say something like "an HTML page which MUST be viewable in a viewport whose dimensions are…", if you want to avoid the term iframe…

thadguidry: Yes would be better to rephrase like that I think.

fsteeg (approved):
Great to see so much text in the spec removed (`+96 −246`)!

Made one small inline suggestion for a missing removal.

Approving this to keep focused on the pre-determined routes vs. announcing them, but since this removes the optional GET routes and changes the way the main `reconcile` and `extend` routes work, this might be the right point to reconsider if we really want to use POST here. In particular if our goal is to get closer to REST principles, I'd think that `reconcile` and `extend` should be GET routes, since they only query data and don't change any data on the server.

(BTW there seems to be an upcoming new verb for 'GET with a body', [currently called SEARCH](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-01.html).)

wetneb:
Thanks, good catch!

For GET vs POST, I guess we are still limited by the fact that:
* passing GET parameters in the URL is subject to length limits
* passing GET parameters in the body is a pretty uncommon / weird thing to do, which might require contortions depending on the framework used for the server or client

But I get that it might also be weird to use POST in some web frameworks, typically those which enforce some sort of CSRF protection on such endpoints (which would need to be disabled for our purposes).

thadguidry:
I'd prefer that we stay with HTTP POST as we have defined in the latest draft.

This would allow a bit more flexibility in the kinds of entities that could potentially be reconciled.  For instance, I wonder if it should be optional for a service to use `multipart/form-data` for binary data and not only `application/x-www-form-urlencoded` but haven't thought about it much beyond use cases of machine learning with file data such as images or sound clips with associated metadata.

acka47:
> BTW there seems to be an upcoming new verb for 'GET with a body', [currently called SEARCH](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-01.html).

By now it's called `QUERY` method. I think this is the most up-to-date draft: https://www.rfc-editor.org/rfc/internet-drafts/draft-ietf-httpbis-safe-method-w-body-03.html

wetneb:
> This would allow a bit more flexibility in the kinds of entities that could potentially be reconciled. For instance, I wonder if it should be optional for a service to use multipart/form-data for binary data and not only application/x-www-form-urlencoded but haven't thought about it much beyond use cases of machine learning with file data such as images or sound clips with associated metadata.

Actually we want to migrate out of such content-types because they are not as REST-friendly as just `application/json`.

Issue #52, opened 2020-09-21
Multilingual support


wetneb:
At the moment, the names and descriptions represented in the reconciliation queries and responses do not come with any language information.

For natively multilingual data sources (such as Wikidata) it would be convenient if this restriction could be lifted. At the moment, this is handled by offering one reconciliation endpoint per language supported by Wikidata (such as https://wikidata.reconci.link/en/api, https://wikidata.reconci.link/it/api, and so on), but it would be nicer to have a single endpoint which would support all languages directly. (It is complicated to teach users to insert the language code in the URL).

The lack of multilingual support was identified by a reviewer from the [Ontology Matching 2020 workshop](http://om2020.ontologymatching.org/) and earlier by @tfmorris in https://github.com/reconciliation-api/specs/pull/48#issuecomment-665802855.

There are other endpoints which also encode some "constants" in the base URL of the reconciliation service. For instance, the OpenCorporates endpoint does this not for languages but for jurisdictions (letting users match against companies from a single country).
https://api.opencorporates.com/documentation/Open-Refine-Reconciliation-API

So perhaps the right way to address this is to provide a better way for services to receive global configuration options, which would encompass the use cases of both the Wikidata and OpenCorporates endpoints?

VladimirAlexiev:
I posted a separate issue #55 of which this one is an instance.
I think we should allow language preferencing, similar to `Accept-Language` and wikidata `label service`. See https://github.com/w3c/sparql-12/issues/13.

Below are draft requirements:

`language` is a special param with these requirements:
- The language param values should conform to [BCP47](http://tools.ietf.org/html/bcp47) and preferably be selected from the [IANA Language Subtag Registry](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) (or see the Google Sheet [iana-lang-tags](https://docs.google.com/spreadsheets/d/1M1yv9aBUmc-NyCJX69vOLUmH2uIglSwmDwgRgByI1AI/edit) for easier access)
- Language matching should conform to [RFC4647](https://tools.ietf.org/html/rfc4647) Matching of Language Tags, as used in HTTP `Accept-Language` and SPARQL [langMatches](https://www.w3.org/TR/sparql11-query/#func-langMatches). Eg "en" should match a name with lang tag "en-GB".
- The language param can take several lang tags separated with commas, in which case they are interpreted as  preference order.
- The service should prefer matches in the specified language(s) but can also return matches in other languages
[...]

PR #103, opened 2022-12-07
Content on accessibility

fsteeg:
Some content on accessibility, going through the W3C [checklist](https://w3c.github.io/apa/fast/checklist.html), as discussed in our [meeting](https://etherpad.lobid.org/p/reconciliation-november-2022). Two things that I noticed (see also inline TODOs):

- Setting a specific size in the [preview service](https://reconciliation-api.github.io/specs/latest/#preview-service) might conflict with clients wanting to provide scaled interfaces for improving accessibility. Do we even need a size here at all? It seems to me the rendering should be up to the client, and the HTML served should be as presentation-independent as possible (no fixed font sizes etc.). We could then remove the [preview metadata](https://reconciliation-api.github.io/specs/latest/#preview-metadata) altogether. If we do need a size, it seems it should be set by the client, not the service. Like, in a header or query parameter of the [preview request](https://reconciliation-api.github.io/specs/latest/#preview-queries)?
- Unlike all other objects (like entities, types, and properties), [matching features](https://reconciliation-api.github.io/specs/latest/#dfn-matching-feature) have no `name`. To provide a uniform way of displaying useful labels, I think we should add that.

I was considering opening issues for these two points, but since they are motivated by the accessibility content here, maybe it makes sense to work on that as part of this pull request?

(There is also the whole internationalization / multilingual support topic, which is very relevant for accessibility, but we have a separate issue for that: #52).


thadguidry:
From our current specs...
>This section specifies how reconciliation services can provide embeddable HTML previews of their entities, which clients can display in their user interface.
>
>width
>The width in pixels of the viewport where to render an entity preview;
>
>height
>The height in pixels of the same viewport.

Hmm, so I would say that it probably is still useful for a client to know what viewport size to use for a preview viewport.  Or put another way...what is the reasonable dimensions that look adequate for the amount of information that a preview is going to push into a viewport.  Some recon service might have LOTS of metadata and others might just be 1 line/sentence, i.e. a dictionary definition.  You might build a client with Electron or Tauri and then use the width/height information to setup the preview window specifically or alter it given some user accessibility preferences (iPhone give me EXTRA LARGE text please).  In OpenRefine we envisioned that later on we might even have a completely different iframe or browser tab for reconciling against services that have lots of disambiguating information about their entities to push into a preview.

1. Knowing what the original intended size (or tested size) of the minimum viewport needed from the recon service would help?  So, I think that's why we have the width/height there in the specs currently.  But do things get magically cutoff or loss of information through a viewport on a client that did not size enough for the minimums of the service...probably so?

2. or scrollbars should be expected...and we mention that in the specs...that a client may handle a viewport differently and should think about accessibility and at a minimum offer scrollbar support (horizontal/vertical) in case viewport size minimums cannot be met for some reason?

fsteeg:
I've updated this as discussed in our December meeting, see commits above for details.

wetneb (approved):
Really nice, thanks a lot!
I hope we can get rid of the note about internationalization soon with a PR that addresses #52, but it makes sense to merge this as is in the meantime.

fsteeg:
Thank you both! Merging.

PR #108, opened 2023-02-09
Add Accept-Language and Content-Language headers

wetneb:
For #52.

There seems to be interest in #52 to have more than that: also introduce parameters in the queries to specify the language of the data. I am not too sure how useful that is and which form it should take (do you want to specify the language of each string in the query independently? should we follow JSON-LD conventions to represent strings with languages?) so I am leaving this for another PR.

### DIFF
--- a/draft/index.html
+++ b/draft/index.html
@@ -462,7 +462,7 @@ database are instances of this type.<dd>
         </p>
         <p>
           Examples of user-facing text in service responses are: the name of the service and the name of property configuration fields in the manifest, the name and description of entities, types and properties,
-          the contents of the entity preview pages.
+          the contents of the entity preview pages, and the documentation linked in the manifest.
         </p>
       </section>
       <section>
--- a/draft/index.html
+++ b/draft/index.html
@@ -455,6 +455,17 @@ database are instances of this type.<dd>
         </p>
       </section>
       <section>
+        <h3>Language Selection</h3>
+        <p>
+           Following [[RFC9110]], services SHOULD use the <code>Accept-Language</code> header to let clients select the language in which user-facing text is returned.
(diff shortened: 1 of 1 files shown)


latest/index.html
```diff
@@ -455,6 +455,17 @@ <h3>Cross-Origin Access</h3>
...
       </section>
+      <section>
+        <h3>Language Selection</h3>
+        <p>
+           Following [[RFC9110]], services SHOULD use the <code>Accept-Language</code> header to let clients select the language in which user-facing text is returned.
+           They SHOULD use the <code>Content-Language</code> in their responses to expose the language each response is returned in.
+        </p>
+        <p>
+          Examples of user-facing text in service responses are: the name of the service and the name of property configuration fields in the manifest, the name and description of entities, types and properties,
+          the contents of the entity preview pages.
```
tfmorris, line 466: This should probably also control what URL is returned for the `documentation` URL in the manifest (unless some other mechanism is proposed for multilingual documentation).

tfmorris (changes requested):

latest/index.html
wetneb: How about the same mechanism? The documentation URL itself could already react to `Accept-Language`, no? Isn't that how most websites (should) work?

tfmorris: That sounds fine.

fsteeg: So should we add the documentation here, like this:

```suggestion
          the contents of the entity preview pages, and the documentation linked in the manifest.
```

fsteeg (approved):

wetneb:
Given https://github.com/reconciliation-api/specs/issues/52#issuecomment-1558986118 I would propose to remove the `Content-Language` header.

tfmorris (approved):

PR #110, opened 2023-02-09
Remove support for single type in queries

### LINKED ISSUE #109: Simplify type syntax in reconciliation queries
Currently, reconciliation queries can contain "either a single type identifier or a list of type identifiers" according to the specs. That is, both of those queries are valid:
```json
{"query":"foo","type":"Q5"}
{"query":"foo","type":["Q5"]}
```

This is encoded in the JSON schema with a `union` construct.
As pointed out by @fsteeg, this is confusing, especially because the specs mention "an array of types" in the definition and then only have single types without arrays in the examples.

For services, it is annoying to parse this cleanly.

I would therefore propose to drop the first syntax above and require types to be enclosed in an array, even if there is only a single element.

wetneb:
Closes #109

### DIFF
--- a/latest/index.html
+++ b/latest/index.html
@@ -218,6 +218,7 @@ href="https://github.com/OpenRefine/OpenRefine/wiki/Reconciliable-Data-Sources">
           <p>Collection of changes which make the API conform to the REST principles.</p>
           <ul>
             <li>Remove support for flyouts;</li>
+            <li>Remove support for supplying a single type in a reconciliation query without enclosing it in an array;</li>
             <li>The endpoints of various sub-services are no longer configurable, but determined by the root reconciliation endpoint.</li>
           </ul>
         </section>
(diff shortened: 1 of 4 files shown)


tfmorris:
Since this is a breaking change, it would be useful to add a note on the compatibility implications (ie legacy clients will be sending a single type ID)

latest/index.html
```diff
@@ -218,6 +218,7 @@ <h4>This Draft</h4>
           <p>Collection of changes which make the API conform to the REST principles.</p>
           <ul>
             <li>Remove support for flyouts;</li>
+            <li>Remove support for supplying a single type in a reconciliation query without enclosing it in an array;</li>
```
wetneb, line 221: Yes, I have added a line about this here - would you want to see that elsewhere? Or written up differently?

thadguidry:
Won't versions of the spec help a little with determining breakage? I.E. if I don't change the version on my recon service that guarantees compatibility in that version and any clients using that same version?

latest/index.html
tfmorris: "Conform to REST principles" doesn't really draw the eye in the same was as "Incompatible changes" or some such, so I'd recommend a separate section. Ideally, it should be easy for a developer of a client or service to understand the implications of the change as described.

wetneb:
Let's go the other direction as suggested by the discussion in #109.

Issue #112, opened 2023-02-10
Clarity on difference between versions


rogerhyam:
I'm trying to implement a Reconciliation Service and am running into a series of problems mainly related to the differences between 0.1 and 0.2.

There is a major difference in the structure of the query batch between the two versions

In 0.1 "A reconciliation query batch is a set of reconciliation queries indexed by string identifiers." (**an object**)

https://reconciliation-api.github.io/specs/0.1/#dfn-reconciliation-query-batch

In 02. "A reconciliation query batch is an **array** of reconciliation queries"

https://reconciliation-api.github.io/specs/latest/#dfn-reconciliation-query-batch

This is a major breaking change but it isn't listed as such

https://reconciliation-api.github.io/specs/latest/#x0-2

"Initial improvements to the specifications made by our Community Group. Most of them are backwards-compatible, except for the requirement to support CORS for cross-origin access."

Edit: The results are also in entirely different formats!

There is no "Candidates" structure in 0.1 but there is in 0.2

https://wikidata.reconci.link/en/api?queries=%7B%22q0%22%3A%7B%22query%22%3A%22banana%22%7D%7D

This is a documentation issue if nothing else.

From a protocol perspective there is an issue of how a service detects what protocol version it is being asked for. Should it simply test if it is being passed an object or an array? If so that should be documented or different services will do it in different ways. Would it be better to declare in the query what version of the protocol is being used? 

Hardly any of the services in the test bench seem to work and fail with similar issues. It looks like the test bench only works with 0.1? Is that correct?

wetneb:
Very sorry that you are having this problem!

We currently have three versions of the specs online:
* https://reconciliation-api.github.io/specs/0.1/, which is version 0.1
* https://reconciliation-api.github.io/specs/0.2/, which is version 0.2
* https://reconciliation-api.github.io/specs/latest/, which is the latest draft of the upcoming version (to be called 0.3 or something else)

Perhaps we should not use the word `latest` in the URL, as it could be interpreted as the "latest released version". Maybe something else, like `draft`, could indicate that better?
Perhaps it would also be worth adding a very visible disclaimer on the current draft that it is being worked on and should not be used by implementers yet?

rogerhyam:
Thanks.

Yes a disclaimer and a recommendation as to which document implementers should be working with would be very helpful!

[...]

PR #113, opened 2023-02-13
Move current draft to the 'draft' directory instead of 'latest'

wetneb:
For #112


fsteeg:
+1 for moving from `latest` to `draft`.

Considering existing links to `latest`, should we add an HTML redirect from `latest/index.html`?

wetneb:
This PR adds a Javascript redirect from `latest/index.html` to `draft/index.html`. I am not sure how to turn it into an HTML redirect (not sure if this is supported by GitHub Pages?)

Arguably one could also decide to redirect `latest/index.html` to `0.2/index.html`, if we want to prioritize the latest stable version of the docs. No strong feeling about this.

fsteeg:
Ah right, sorry, somehow I missed that, I thought `latest/index.html` was deleted. I've added what I expected as an HTML redirect in https://github.com/reconciliation-api/specs/pull/113/commits/30ac7a6c42f13b0bc971296d7464824bd46a0ecf, which I verified to work with JS disabled locally.

I also like the idea of redirecting `latest` to `0.2`, the latest stable version.

PR #119, opened 2023-03-25
Add schema for data extension property proposal responses

### LINKED ISSUE #118: No JSON schema for data extension property proposal response
In https://reconciliation-api.github.io/specs/0.2/#dfn-data-extension-property-proposal-response, Respec (the spec-writing framework we use) complains that we never refer to the definition of a data extension property proposal response.

We should add a JSON Schema for this response and use the existing mechanism to validate valid and invalid examples against it.

wetneb:
Closes #118.
(To be duplicated in `/0.2` - not included in this PR to keep it concise)

### DIFF
--- a/draft/examples/data-extension-property-proposal/invalid/missing-properties.json
+++ b/draft/examples/data-extension-property-proposal/invalid/missing-properties.json
@@ -1,0 +1,4 @@
+{
+  "limit": 5,
+  "type": "DifferentiatedPerson"
+}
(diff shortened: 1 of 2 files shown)


fsteeg (approved):
Suggested one minor typo fix inline.

Also noticed some descriptions end with `.`, others don't, but that's true for the other schemas too.

wetneb:
Thanks, good catch! Opening another issue to normalize descriptions: #120

PR #123, opened 2023-04-10
Set up redirect for 0.1 and 0.2 specs

### LINKED ISSUE #122: Redirect to the final version of the 0.1 specs
Since I have taken the steps to publish the 0.1 specs on the W3C website, we have a couple of problems at https://reconciliation-api.github.io/specs/0.1/:
* The link at "This version" points to a non-existing page because the date is automatically picked as today's date. The correct link is the "Last published version" underneath
* The page takes quite some time to load with Respec trying to fetch references dynamically (and failing, somehow?)

Therefore I would propose to remove this page and replace it by a redirect to https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.1-20230321/.
We could move `0.1/index.html` to something like `0.1/source.html` to keep the source of the published spec around (in case we want to make minor corrections later, which I think we should still be able to publish at the same URL on `w3.org`).

wetneb:
Closes #122


0.1/index.html
```diff
@@ -1,855 +1,14 @@
-<!DOCTYPE html>
+<!doctype html>
```
fsteeg, line 1: Is the lowercasing of the `DOCTYPE` intentional? It seems to be valid HTML, but some related tech seems to expect uppercase (see https://stackoverflow.com/a/23946325).

fsteeg (approved):
Looks good, added one inline comment.

Issue #125, opened 2023-05-09
I18n checklist: language


fsteeg:
This is a self-review checklist based on the <a href="https://www.w3.org/TR/international-specs/#language_checklist">Internationalization Best Practices for Spec Developers</a>.

(All lines seemed relevant for us here, I suggest we ~~strikeout~~ things that turn out not to be.)

See also issue #52, PR #108 and notes at https://etherpad.lobid.org/p/reconciliation-i18n.

Check marks indicate requirements that are fulfilled by this spec. Add comments to describe conformance.

### Language basics
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> It should be possible to associate a language with any piece of <a data-type="dfn" href="https://www.w3.org/TR/i18n-glossary/#dfn-localizable-text">localizable text</a> or <a data-type="dfn" href="https://www.w3.org/TR/i18n-glossary/#dfn-natural-language">natural language</a> content. <a href="https://www.w3.org/TR/international-specs/#lang_basics_1" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> Where possible, there should  be a way to label <a data-type="dfn" href="https://www.w3.org/TR/i18n-glossary/#dfn-natural-language">natural language</a> changes in inline text. <a href="https://www.w3.org/TR/international-specs/#lang_basics_inline" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> Consider whether it is useful to express the <a href="https://w3c.github.io/bp-i18n-specdev/#sec_lang_meta">intended linguistic audience</a> of a resource, in addition to specifying the language used for <a href="https://w3c.github.io/bp-i18n-specdev/#sec_text_processing_lang">text processing</a>. <a href="https://www.w3.org/TR/international-specs/#lang_basics_meta" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_
[...]

PR #129, opened 2023-05-17
Add section on text-processing language

fsteeg:
As discussed in our [last meeting](https://etherpad.lobid.org/p/reconciliation-may-2023#L60), this PR addresses the text-processing language in our spec, based on the discussion and the proposal in https://github.com/reconciliation-api/specs/issues/52#issuecomment-1424411630. See also #125 and https://www.w3.org/TR/international-specs/#sec_text_processing_lang.


wetneb (approved):
It looks astoundingly simple and natural!

Issue #128, opened 2023-05-11
Add an additional property match context to the features


thadguidry:
1. It would be useful to provide clients and for clients to filter on a match context in order to have more context on a candidate's score or condition.  This should be a simple string to ease development pain for services that need something simple to start with.
2. Some services might even want to use different `feature_view` 's depending on a `serviceVersion` or even `schemaSpace`?  But not sure about that case myself directly.
3. A standard client could also query 2 alternative services by a provider. (different service URLs and service manifests at each URL) and might produce different context and scores for the same data, type, properties supplied in the query.

Regardless, it is sometimes useful for clients to know the match context of candidates (or features of candidates) from a recon process against a service, if the service decides to provide a bit more context or information about a match score, an entity itself, or types or properties that were used or not used, etc. etc.

A match `context` provides a simple means of returning extra metadata or subdata about a match overall, and not necessarily about an individual feature, although it could also say much about that as well since the value type of `context` would simply be a `String`.

Example 1:

```json
{
  "id": "1117582299",
  "name": "Urbaniak, Hans-Eberhard",
  "score": 85.71888,
  "context": "used v1.6 with weighting based on KL divergence" 
  "features": [
    {
      "id": "name_tfidf",
      "name": "TF-IDF score for the entity name",
      "value": 378.239
    },
 ```

Example 2:
 
 ```json
{
  "id": "12345",
  "name": "generic drug label",
  "score": 55,
  "features": [
    {
      "id": "name_generic",
      "name": "baseline score for the label",
      "value": 133
      "context": "non-LSI, 1 matched broader type: generic" 
    },
 ```
 
 I envision that `context` might be most useful at the candidate level in the first example, but perhaps also in `features` as in Example 2?
 
 @wetneb let me know which parts above are unclear and I can update this.  This is simple for a reason, and does not directly address larger client-service feedback loops, but a step in that direction for broader applicability and uptake by service providers. (hopefully)

fsteeg:
[...]

Issue #106, opened 2023-01-12
Reformulate the `query` parameter as one of the `properties` in reconciliation queries


wetneb:
The name to be matched in a reconciliation query is supplied in a dedicated field, the `query` one:
```json
{
    "query": "Jane Doe",
    "properties": [
         {"pid":"date_of_birth", "v": "1967-07-06"}
    ]
}
```

On various occasions, it has come up that it could make sense to reformulate this as a special property, such as:
```json
{
    "properties": [
         {"pid":"name", "v": "Jane Doe"},
         {"pid":"date_of_birth", "v": "1967-07-06"}
    ]
}
```

This would be useful for #101, or could also have provided a sensible solution to #27.

That being said, it does require introducing a sort of special property, which could conflict with the naming of other properties supported by the service. It is also a significant breaking change.

thadguidry:
Why special at all?  Entities have properties.  Name is just one of many potential properties.
And in that case, we could think to label the property "entityname" and likely avoid conflicts?

PR #131, opened 2023-07-13
Optional details for matching properties in reconciliation queries

### LINKED ISSUE #88: Allow more flexible filter constraint handling of property arrays
In order to fully accommodate `features` and provide better accuracy computing `score` to clients against a set of optional or required filter constraints of properties that a user is requesting.

We should allow multiple property arrays and within each array the combining behaviors that were typical in the past (Freebase) with recon services (See the old [Freebase advanced filtering](https://developers.google.com/freebase/v1/search-overview#advanced-filtering) and [search cookbook](https://developers.google.com/freebase/v1/search-cookbook#textual-constraints) )
) 
* AND (JSON Schema - `allOf`)
* OR  (JSON Schema - `anyOf`)
* NOT (JSON Schema - `not`)
* SHOULD

Those would also avoid the need for Union types, since we could construct something like `(or P104:"box" P104:"ball")`

And individual properties could be filtered similarly with suffixes as in Prisma backend with GraphQL such as:

```
_gt (greater than)
_lt (less than)
_gte (greater than or equal to)
_lte (less than or equal to)
_in (equal to)
_not_in (not equal to)
```

Example in GraphQL
I.E. return all persons' details only if their age is greater than 18:

```graphql
query {
  persons(where: {age_gt: 18}) {
    firstName
    lastName
    age
  }
}
```

In the OpenRefine reconciliation service issue # https://github.com/wetneb/openrefine-wikibase/issues/141#issuecomment-1170311259 where I commented that currently it seems it is impossible to use the current reconciliation api standard to form a filtering query such as this example:

```
?query="William Albert Ablett"&filter=(any P735:"Albert" P735:"John" (and P734:"Ablett"))"
```

### QUESTION 1. 
In the current spec, we do not mention the cardinality of `properties`, but it seems we implied that each query object can only have 1 `properties` array?

If using the current spec, there is no way to do so with a single query while also using the `properties[]` and `type_strict`, since supplying multiple `properties[]` breaks the implicit relationship of `type_strict`. (which `type_strict` attaches to which `properties[]` array?)  We certainly could move `type_strict` inside of `properties[]` but this is a major change to the API and we have additional questions similarly raised in issue #60   

### QUESTION 2.
So then It seems we would need at least 2 queries, which defeats the extra narrowed filtering that could be accomplished in a single request?

match entities named "William Albert Ablett"
that have either "Albert" or "John" as a first name property
and "Ablett" as a last name property

```json
{
  "q0": {
    "query": "William Albert Ablett",
    "type": "Q5",
    "limit": 5,
    "properties": [
      {
        "pid": "firstName",
        "v": ["Albert","John"]
      }
    ],
    "type_strict": "any",
    "properties": [
      {
        "pid": "lastName",
        "v": "Albert"
      }
    ],
    "type_strict": "and"
 }
}
```

### QUESTION 3.
If desired by some devs, how could we make it easier to allow some form of Union types, such as in GraphQL, and what might that best look like?  https://graphql.org/learn/schema/#union-types

### LINKED ISSUE #101: Add ability to mark reconciliation property as required (used for candidate retrieval)
Currently, [reconciliation queries](https://reconciliation-api.github.io/specs/latest/#structure-of-a-reconciliation-query) consist in:
* a name to search for (`query`)
* a list of properties to refine the search (`properties`)

Each of those properties comes with a property identifier (`pid`) and a value (`v`).

This lets clients refine the search for reconciliation candidates, by providing additional data to be taken into account.

Most reconciliation pipelines feature two separate steps:
* candidate retrieval: extracting a set of candidates which are likely to match the query. This is generally done using indices (also called "blocking" techniques in the literature - basically anything that lets the system avoid doing a full scan of the database but instead locate the candidates efficiently).
* candidate scoring: this smaller set of candidates is processed one by one, comparing attributes of the retrieved entities with the query, to come up with a matching score

At the moment it is up to the services to determine which parts of a reconciliation query they use for candidate retrieval or matching. This is often dictated by the structure of the database they are exposing, as indices are expensive to construct and maintain.

However, in certain cases it would be helpful to give the user some control over which parts of the query should be used at the retrieval stage.

Consider for instance a batch of reconciliation queries for cities, consisting of a name (in the `query` field) and of geographical coordinates (in some property supplied in the `properties` field). Potentially, the reconciliation service could have two indices at its disposal:
* a full text search index on the entity names, which could be used to retrieve candidates using the supplied names
* a geographical index on the entity coordinates, which could be used to retrieve candidates using the supplied coordinates.

With their knowledge of the quality of the data to reconcile (and of the structure of the authority database), the user could either decide:
1. that names are more reliable, hence they should be used for candidate retrieval, and the coordinates should only be used as tie-breakers between namesakes at the scoring phase
2. or that coordinates are more reliable, so they should be used for retrieval, and names are then used for scoring in the second step

At the moment our API does not let the user specify this sort of information. I think it would be worth introducing a syntax for that, for instance (random JSON syntax made up without putting a lot of thought into it):

Scenario 1 would be
```json
{
    "query": "Cambridge",
    "query_hint": "retrieval",
    "properties": [
         {
             "pid": "coordinates",
             "v": "52.205278,0.119167",
             "hint": "scoring"
         }
    ]
}
```
and scenario would be:
```json
{
    "query": "Cambridge",
    "query_hint": "scoring"
    "properties": [
         {
             "pid": "coordinates",
             "v": "52.205278,0.119167",
             "hint": "retrieval"
         }
    ]
}
```

Because we cannot reasonably expect services to have indices on every property they expose, I think this feature should be a `SHOULD` or `MAY` and not a `MUST` in the specs (hence the proposed syntax of `hint`, that the services can decide to follow or not).

## See also

- Related feature requests in OpenRefine: https://github.com/OpenRefine/OpenRefine/issues/150 and to some extent https://github.com/OpenRefine/OpenRefine/issues/2433

### LINKED ISSUE #105: Supplying multiple properties with the same pid in a reconciliation query
The specs currently do not forbid the following query:
```json
{
    "query": "John Doe",
    "properties": [
         {"pid": "P123", "v": "first value"},
         {"pid": "P123", "v": "second value"}
     ]
}
```
In other words, we do not say whether the value of the `pid` field is expected to be unique in the `properties` array.

It is unclear to me if such a query should instead be reformulated as:
```json
{
    "query": "John Doe",
    "properties": [
         {"pid": "P123", "v": ["first value", "second value"]}
     ]
}
```
I guess services could interpret both queries differently for matching purposes, perhaps giving a use to both queries?
In any case, it would be worth talking about that in the specs.

Brought up by @paulgirard in today's OpenRefine contributors meetup.

### LINKED ISSUE #114: Date comparison needs an absolute settings to compute relative diff score
When testing matches on date properties, the recon service needs to output a 0-100 score from a date comparison.
If scoring 100 is clear (perfect match), the continuous score domain between 0 an 99.99 is not.
It demands to define a scale of time which would transform a time difference into a 0-99.99 float number.
This needs a decision from the user as it depends on the use case.
Users aligning items from a long-run history study spanning millenniums would not require the same scale as contemporary event date for instance.

To illustrate here is a commit I issued to adapt the wikidata recon api date scoring to a personal use case:
https://gitlab.com/ouestware/openrefine-wikibase/-/commit/ee2bf57842b2d0dd00cf7e3f916461c03aabb164
(I think the scoring system I drafted should be simplified to express score as a function of time difference)

It might worth letting the user define the date precision scale in the query.

One way could be to define from which time duration (express as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Durations)) difference score should be 0.
Which is equivalent to define the time approximation span in which candidates dates must be to be scored as potential candidates.
But there are many more ways to solve this issue I am sure?

Laziness disclosure: I haven't checked if such a settings already exist...

I hope this reflexion could help future improvements of the spec.

fsteeg:
This adds three optional fields to properties in reconciliation queries. It's based on our discussions in [recent meetings](https://etherpad.lobid.org/p/reconciliation-meetings), in particular the suggestion in https://github.com/reconciliation-api/specs/issues/105#issuecomment-1463754747, adding an additional field to address #114, based on https://github.com/reconciliation-api/specs/issues/128#issuecomment-1592778805.

Changes from the linked discussions:

- Replaced `mode=must|should` with `required=true|false` (we discussed `filter` as an option in our last meeting, but I felt `required` was clearer since filter can be used both meaning "accept" and "reject")
- Renamed `match=any|all|none` to `match_quantifier=any|all|none`
- Added `match_qualifier=<string>` to address use cases like #114, which came up in otherwise unrelated https://github.com/reconciliation-api/specs/issues/128#issuecomment-1592778805

This is related to multiple issues, which I'd hope this could resolve:

- Might resolve #88 via `match_quantifier`
- Might resolve #101 via `required`
- Might resolve #105 by supporting multiple property statements with different configurations, as well as configuration of multiple values in one property statement via "any"/"all"/"none" (`match_quantifier`)
- Might resolve #114 via `match_qualifier` and the mention of the EDTF specification
- Might resolve parts of #106, which maybe could have been resolved before, since `query` is optional already, but with this we have precise control over the behavior of the property replacing the `query`

### DIFF
--- a/draft/examples/reconciliation-query-batch/valid/example-full.json
+++ b/draft/examples/reconciliation-query-batch/valid/example-full.json
@@ -7,11 +7,17 @@
       "properties": [
         {
           "pid": "professionOrOccupation",
-          "v": "Politik*"
+          "v": "Politik*",
+          "required": false,
+          "match_quantifier": "any",
+          "match_qualifier": "WildcardMatch"
         },
         {
           "pid": "affiliation",
-          "v": "http://d-nb.info/gnd/2022139-3"
+          "v": "http://d-nb.info/gnd/2022139-3",
+          "required": false,
+          "match_quantifier": "any",
+          "match_qualifier": "ExactMatch"
         }
       ]
     },
@@ -22,11 +28,17 @@
       "properties": [
         {
           "pid": "professionOrOccupation",
-          "v": "Politik*"
+          "v": "Politik*",
+          "required": false,
+          "match_quantifier": "any",
+          "match_qualifier": "WildcardMatch"
         },
         {
           "pid": "affiliation",
-          "v": "http://d-nb.info/gnd/2022139-3"
+          "v": "http://d-nb.info/gnd/2022139-3",
+          "required": false,
+          "match_quantifier": "any",
+          "match_qualifier": "ExactMatch"
         }
       ]
     }
--- a/draft/examples/suggest-properties-response/valid/example.json
+++ b/draft/examples/suggest-properties-response/valid/example.json
@@ -3,12 +3,17 @@
     {
       "name": "coordinate location",
       "description": "geocoordinates of the subject. For Earth, please note that only WGS84 coordinating system is supported at the moment",
-      "id": "P625"
+      "id": "P625",
+      "match_qualifiers": [
+        {"id": "ExactMatch", "name": "Exact match of the coordinates"},
+        {"id": "DecimalPlaces-N", "name": "Match the coordinates with a precision of N decimal places"}
+      ]
     },
     {
       "name": "place of birth",
       "description": "most specific known (e.g. city instead of country, or hospital instead of city) birth location of a person, animal or fictional character",
-      "id": "P19"
+      "id": "P19",
+      "match_qualifiers": ["schema:containsPlace", "schema:containedInPlace"]
     },
     {
       "name": "located in time zone",
--- a/draft/examples/suggest-properties-response/valid/example.json
+++ b/draft/examples/suggest-properties-response/valid/example.json
@@ -13,7 +13,10 @@
       "name": "place of birth",
       "description": "most specific known (e.g. city instead of country, or hospital instead of city) birth location of a person, animal or fictional character",
       "id": "P19",
-      "match_qualifiers": ["schema:containsPlace", "schema:containedInPlace"]
+      "match_qualifiers": [
+        {"id": "schema:containsPlace", "name": "Containment relation between a place and another that it contains"},
+        {"id": "schema:containedInPlace", "name": "Containment relation between a place and another that contains it"}
+      ]
     },
     {
       "name": "located in time zone",
--- a/draft/index.html
+++ b/draft/index.html
@@ -96,6 +96,11 @@
             "publisher": "IETF",
             "href": "https://www.rfc-editor.org/rfc/bcp/bcp47.txt"
           },
+          "EDTF": {
+            "title": "Extended Date/Time Format Specification (part of ISO 8601:2019)",
+            "publisher": "Library of Congress / International Organization for Standardization",
+            "href": "https://www.loc.gov/standards/datetime/"
+          },
         }
       };
     </script>
@@ -517,12 +522,26 @@ database are instances of this type.<dd>
 	    <dt><code>limit</code></dt>
             <dd>A limit on the number of candidates to return, which must be a positive integer;</dd>
 	    <dt><code>properties</code></dt>
-	    <dd>An array of objects, where each object maps a <a href='#properties'>property</a> identifier (in the <code>pid</code> field) 
-        to one or more <a>property values</a> (in the <code>v</code> field). These are used to further filter the set of candidates (similar to a WHERE clause in SQL),
-		by allowing clients to specify other attributes of entities that should match, beyond their name in the <code>query</code> field.
-		How reconciliation services handle this further restriction ("must match all properties" or "should match some") and how it affects the score, is up to the service.
-		A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties;</dd>
-          </dl>
+        <dd><p>An array of objects, where each object maps a <a href='#properties'>property</a> identifier (in the <code>pid</code> field) to one or more <a>property values</a> (in the <code>v</code> field).
+        These are used to further refine the list of candidates by allowing clients to specify other attributes of entities, beyond their name in the <code>query</code> field.
+        A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties.</p>
+        <p>In addition to <code>pid</code> and <code>v</code>, services MAY support the following optional fields that allow clients to specify the effect of each property on the resulting list of candidates.
+          If these fields are omitted, the exact behavior ("must match all", "should match some", etc.) is up to the service.</p>
+        <dl>
+          <dt><code>required</code></dt>
+          <dd>A boolean indicating if a match for the property is required for an entity to enter the list of candidates (i.e. acting like a filter or a WHERE clause in SQL)
+            or optional (i.e. only effecting the entity's rank in the list of candidates);</dd>
+          <dt><code>match_quantifier</code></dt>
+          <dd>A string to indicate which of the values in <code>v</code> to match. MUST be "any" (equivalent to boolean OR), "all" (equivalent to boolean AND), or "none" (equivalent to boolean NOT);</dd>
+          <dt><code>match_qualifier</code></dt>
+          <dd>A string to indicate how to match the values in <code>v</code>.
+            This can be used for general matching relations like "skos:exactMatch", "skos:closeMatch", etc. or for specific features like spatial matching with geo data
+            (e.g. containment search with "schema:containsPlace" etc.) or custom matching on date fields (e.g. services supporting the [[EDTF]] specification could use "EDTF:Level-0" etc.
+            To allow discovery of supported qualifiers by clients, services that support <code>match_qualifier</code> SHOULD return a <a href='#suggest-responses'>suggest response</a> with supported qualifiers
+            for a property at <code>/suggest/property/match_qualifier/&lt;pid&gt;</code> (relative to the service root).</p></dd>
+        </dl>
+        </dd>
+        </dl>
         </p>
         <p>
           A <dfn>reconciliation query batch</dfn> is an array of <a>reconciliation queries</a>.
--- a/draft/index.html
+++ b/draft/index.html
@@ -537,8 +537,8 @@ database are instances of this type.<dd>
           <dd>A string to indicate how to match the values in <code>v</code>.
             This can be used for general matching relations like "skos:exactMatch", "skos:closeMatch", etc. or for specific features like spatial matching with geo data
             (e.g. containment search with "schema:containsPlace" etc.) or custom matching on date fields (e.g. services supporting the [[EDTF]] specification could use "EDTF:Level-0" etc.
-            To allow discovery of supported qualifiers by clients, services that support <code>match_qualifier</code> SHOULD return a <a href='#suggest-responses'>suggest response</a> with supported qualifiers
-            for a property at <code>/suggest/property/match_qualifier/&lt;pid&gt;</code> (relative to the service root).</p></dd>
+            To allow discovery of supported qualifiers by clients, services that support <code>match_qualifier</code> SHOULD return the supported <code>match_qualifiers</code> for each property
+            in their property <a href='#suggest-responses'>suggest responses</a>.</p></dd>
         </dl>
         </dd>
         </dl>
@@ -757,7 +757,9 @@ in the <code>score</code> field). By exposing individual features in their respo
               <dt><code>description</code></dt>
               <dd>An optional description which can be provided to disambiguate namesakes, providing more context. This could for instance be displayed underneath the <code>name</code>;</dd>
 	      <dt><code>notable</code></dt>
-	      <dd>When suggesting entities only, this field can be used to supply some important types (not necessarily all types) of the suggested entity. The value must be an array of either type identifiers (as strings) or type objects, containing a <code>id</code> and <code>name</code> field which represent the type.</dd>
+	      <dd>When suggesting entities only, this field can be used to supply some important types (not necessarily all types) of the suggested entity. The value must be an array of either type identifiers (as strings) or type objects, containing an <code>id</code> and <code>name</code> field which represent the type.</dd>
+              <dt><code>match_qualifiers</code></dt>
+              <dd>When suggesting properties only, an optional array of either identifiers (as strings) or objects containing an <code>id</code> and <code>name</code> field, which represent the property's <code>match_qualifiers</code> supported in <a>reconciliation queries</a>.</dd>
             </dl>
             </dd>
           </dl>
--- a/draft/index.html
+++ b/draft/index.html
@@ -759,7 +759,7 @@ in the <code>score</code> field). By exposing individual features in their respo
 	      <dt><code>notable</code></dt>
 	      <dd>When suggesting entities only, this field can be used to supply some important types (not necessarily all types) of the suggested entity. The value must be an array of either type identifiers (as strings) or type objects, containing an <code>id</code> and <code>name</code> field which represent the type.</dd>
               <dt><code>match_qualifiers</code></dt>
-              <dd>When suggesting properties only, an optional array of either identifiers (as strings) or objects containing an <code>id</code> and <code>name</code> field, which represent the property's <code>match_qualifiers</code> supported in <a>reconciliation queries</a>.</dd>
+              <dd>When suggesting properties only, an optional array of objects, each containing an <code>id</code> and <code>name</code> field, which represent the property's <code>match_qualifiers</code> supported in <a>reconciliation queries</a>.</dd>
             </dl>
             </dd>
           </dl>
--- a/draft/schemas/reconciliation-query-batch.json
+++ b/draft/schemas/reconciliation-query-batch.json
@@ -82,6 +82,23 @@
                       }
                     }
                   ]
+                },
+                "required": {
+                  "type": "boolean",
+                  "description": "A boolean indicating if a match for the property is required for an entity to enter the list of candidates"
+                },
+                "match_quantifier": {
+                  "type": "string",
+                  "description": "A string to indicate which of the values in v to match",
+                  "enum": [
+                    "any",
+                    "all",
+                    "none"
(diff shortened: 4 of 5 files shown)


draft/examples/reconciliation-query-batch/valid/example-full.json
```diff
@@ -7,11 +7,17 @@
       "properties": [
         {
           "pid": "professionOrOccupation",
-          "v": "Politik*"
+          "v": "Politik*",
```
tfmorris, line 10: Lexical matching seems like a stretch here. I think it would be much more likely for the user to want any sub-class of a given profession.

draft/index.html
```diff
@@ -517,12 +522,26 @@ <h3>Structure of a Reconciliation Query</h3>
...
+            or optional (i.e. only effecting the entity's rank in the list of candidates);</dd>
+          <dt><code>match_quantifier</code></dt>
+          <dd>A string to indicate which of the values in <code>v</code> to match. MUST be "any" (equivalent to boolean OR), "all" (equivalent to boolean AND), or "none" (equivalent to boolean NOT);</dd>
+          <dt><code>match_qualifier</code></dt>
+          <dd>A string to indicate how to match the values in <code>v</code>.
+            This can be used for general matching relations like "skos:exactMatch", "skos:closeMatch", etc. or for specific features like spatial matching with geo data
+            (e.g. containment search with "schema:containsPlace" etc.) or custom matching on date fields (e.g. services supporting the [[EDTF]] specification could use "EDTF:Level-0" etc.
+            To allow discovery of supported qualifiers by clients, services that support <code>match_qualifier</code> SHOULD return a <a href='#suggest-responses'>suggest response</a> with supported qualifiers
+            for a property at <code>/suggest/property/match_qualifier/&lt;pid&gt;</code> (relative to the service root).</p></dd>
```
tfmorris, line 541: Seems like this list should be small enough that it could just be returned with the response to the original property suggest query, although it probably needs descriptions too to help the user figure out what the heck they mean.

draft/index.html
```diff
@@ -517,12 +522,26 @@ <h3>Structure of a Reconciliation Query</h3>
...
+        A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties.</p>
+        <p>In addition to <code>pid</code> and <code>v</code>, services MAY support the following optional fields that allow clients to specify the effect of each property on the resulting list of candidates.
+          If these fields are omitted, the exact behavior ("must match all", "should match some", etc.) is up to the service.</p>
+        <dl>
+          <dt><code>required</code></dt>
+          <dd>A boolean indicating if a match for the property is required for an entity to enter the list of candidates (i.e. acting like a filter or a WHERE clause in SQL)
+            or optional (i.e. only effecting the entity's rank in the list of candidates);</dd>
+          <dt><code>match_quantifier</code></dt>
+          <dd>A string to indicate which of the values in <code>v</code> to match. MUST be "any" (equivalent to boolean OR), "all" (equivalent to boolean AND), or "none" (equivalent to boolean NOT);</dd>
+          <dt><code>match_qualifier</code></dt>
```
tfmorris, line 536: This feels like a solution in search of a problem. Are there use cases described which need this? I'm tempted to just play the YAGNI card here.

draft/index.html
```diff
@@ -517,12 +522,26 @@ <h3>Structure of a Reconciliation Query</h3>
...
-		A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties;</dd>
-          </dl>
+        <dd><p>An array of objects, where each object maps a <a href='#properties'>property</a> identifier (in the <code>pid</code> field) to one or more <a>property values</a> (in the <code>v</code> field).
+        These are used to further refine the list of candidates by allowing clients to specify other attributes of entities, beyond their name in the <code>query</code> field.
+        A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties.</p>
+        <p>In addition to <code>pid</code> and <code>v</code>, services MAY support the following optional fields that allow clients to specify the effect of each property on the resulting list of candidates.
+          If these fields are omitted, the exact behavior ("must match all", "should match some", etc.) is up to the service.</p>
+        <dl>
+          <dt><code>required</code></dt>
```
tfmorris, line 531: I recognize that renaming things is one of the pounds of flesh required by the standardization process, but hopefully the revision notes will include a mapping from the old (`match=`) to the new (`required=`).

tfmorris (commented):
It feels like unneeded complexity to me. One of the things that I like about the [IETF process](https://www.ietf.org/standards/process/) is that they require prior implementation and testing. This demonstrates that a) it's implementable and b) there's at least one real world use case.

It feels like this spec is accumulating more and more words/features without a commensurate increase in real world testing experience (unless there's implementation testing going on behind the scene that I'm missing).

But it's all just intuition / gut feeling...

draft/index.html
wetneb: I think what @fsteeg meant by "renaming" is a renaming from the other solutions proposed in the corresponding issues tackled by this PR, not from any parameter currently existing in the specs. At least I am not aware where this `mode` parameter would currently be used in the specs (even in the `draft` version).

fsteeg: Yes, sorry, my wording was confusing here (Antonin understood it the same way in our meeting yesterday). This PR only adds 3 new, optional fields.

draft/index.html
fsteeg: So my attempt here with `match_qualifier`, in particular combined with the mention of EDTF (brought into the discussion by @thadguidry), was to address @paulgirard's use case from https://github.com/reconciliation-api/specs/issues/114 (which we also discussed in our [April meeting](https://etherpad.lobid.org/p/reconciliation-april-2023)). Paul, would this proposal here help with your use case?

draft/index.html
fsteeg: Yes, that would be an option too. My line of thought was that since this is a separate selection step in the client (first select the property, then select the qualifier), clients could reuse existing suggest endpoint logic (do the same lookup as for the property itself, but against a different endpoint), instead of building custom UI. But that may be a faulty assumption, since I'm looking at it more from the service side.

draft/examples/reconciliation-query-batch/valid/example-full.json
fsteeg: Yes, true, that is a bit odd. This example stems from the GND data that our service exposes, where we don't have a common entity for female and male politicians, so this is a way to cover both. However, that line is not actually added with this PR, it's only green because of the comma, this has been the example before (not to say this should not be changed, but maybe not in this PR, in particular I was happy that this actually provided another usage of `match_qualifier`, using `WildcardMatch`, thus formalizing something that the service was doing implicitly).

fsteeg:
Thanks a lot for your review @tfmorris! I replied inline to your specific comments.

> It feels like this spec is accumulating more and more words/features without a commensurate increase in real world testing experience (unless there's implementation testing going on behind the scene that I'm missing). But it's all just intuition / gut feeling...

I got to say I do share that gut feeling! What I actually tried to do here was to condense all these different suggestions from the 5 linked issues into a fairly compact, non-breaking proposal (optional fields and retaining the current behavior if omitted) that addresses actual use cases.

There is one use case I forgot to mention: https://github.com/OpenRefine/OpenRefine/issues/5615 could be implemented using `required` properties with `match_quantifiers` instead of the type filter.

> One of the things that I like about the [IETF process](https://www.ietf.org/standards/process/) is that they require prior implementation and testing. This demonstrates that a) it's implementable and b) there's at least one real world use case.

Yes, I agree, and it's what I like so much about working on this spec, that it is based on years of proven functionality, and not "made up". I very much want to keep it that way too! After we addressed any obvious issues with the approach proposed in this PR, we should implement it in the testbench and our service to see if it actually works.

draft/index.html
paulgirard: The `match_quantifier` justifies the possibility to add the same property id multiple time is the same request to create compound queries. It's an answer to #105  I think that the possibility to mention "any" or "all" is very important to tweak candidates selection.
The `match_qualifier` is definitely an very elegant answer to the date issue #114. I think this proposal is smart by letting recon services expose what kind of magic they propose and let any query decide to use it or not.

paulgirard (approved):
This proposal is a nice solution to the recent issues about the difficulty of selecting candidates.
We already discussed the complexity issue and to whether put to put the candidate selection decision in server or client side. The conclusion of this proposal feels balanced to me in this regards. One one hand we unfortunately can't move everything on client side leaving weird fuzzy matching to client as candidates pre-selection requires that kind of complexity. On the other those new parameters are optional leaving recon services developers/maintainers decide what to do.
Thus I feel like this would help recon services to grow in diversity and quality.

draft/index.html
tfmorris: Yes, I was confused. It looks like the old syntax was actually `filter=all,any,not,should`, along with `exact=true,false` https://developers.google.com/freebase/v1/search but since there's been such a big gap in time between implementations, it's probably not a big deal.

draft/examples/reconciliation-query-batch/valid/example-full.json
tfmorris: "Things not strings." If this can't be fixed in the GND model, I'd rather see the reconciliation service work around it by introducing its own fake super type than supporting lexical wildcarding on properties.

But since this wasn't a change introduced by this pull request, I'm happy deferring fixing it.

wetneb (approved):
Looks good to me.

I think I disagree that this solves #106 though. After this PR, the `query` field still has a sort of special status, and there is no standard way to replace it by a property (since clients cannot know which `pid` to use). But I think this PR makes it all the more interesting to solve #106, since the new settings you introduce cannot be applied to the `query` itself.

I agree with @tfmorris that we've been introducing quite a few things in the specs and the implementations are lagging behind. I'm keen to try things out in OpenRefine and some services but I have more urgent things on my plate in OpenRefine at the moment. Also it's a requirement for the standardization process at W3C that there are at least 2 independent implementations so there is no risk that we adopt something completely made up and untested.

thadguidry:
No matter what syntax, we need to keep the ability for users to always optionally provide constraints (filters and property matching criteria) at query time, and not depend on services to have some fancy logic to figure things out the clients wants or wishes to return an adequate response.  So I always prefer syntax that keeps more control for users.

draft/index.html
fsteeg: Given the feedback by @tfmorris here and by @wetneb on the UI differences [in our last meeting](https://etherpad.lobid.org/p/reconciliation-august-2023#L28) ("the set is small, and we'd like to see all options, without typing"), I've moved the `match_qualifiers` discovery into the property suggest response: https://github.com/reconciliation-api/specs/pull/131/commits/b486232.

draft/examples/suggest-properties-response/valid/example.json
```diff
@@ -3,12 +3,17 @@
...
+        {"id": "ExactMatch", "name": "Exact match of the coordinates"},
+        {"id": "DecimalPlaces-N", "name": "Match the coordinates with a precision of N decimal places"}
+      ]
     },
     {
       "name": "place of birth",
       "description": "most specific known (e.g. city instead of country, or hospital instead of city) birth location of a person, animal or fictional character",
-      "id": "P19"
+      "id": "P19",
+      "match_qualifiers": ["schema:containsPlace", "schema:containedInPlace"]
```
wetneb, line 16: I think we should settle on using only one format for the `match_qualifiers` field, otherwise it's painful to parse for clients. I'd go for the format with objects (which makes it possible to display human-readable info and is more extensible).

Also, the corresponding JSON schema should be updated accordingly (now that we are close to merging this, I think)

fsteeg: Implemented in https://github.com/reconciliation-api/specs/pull/131/commits/c3c220a535805cac8ac75e6a90b503a5af8d7431 & https://github.com/reconciliation-api/specs/pull/131/commits/fb606689307067b26ed0fdbcd232130aa71d6599.

wetneb (approved):
Looks good to me!

draft/index.html
```diff
@@ -517,12 +522,26 @@ <h3>Structure of a Reconciliation Query</h3>
...
+            or optional (i.e. only effecting the entity's rank in the list of candidates);</dd>
+          <dt><code>match_quantifier</code></dt>
+          <dd>A string to indicate which of the values in <code>v</code> to match. MUST be "any" (equivalent to boolean OR), "all" (equivalent to boolean AND), or "none" (equivalent to boolean NOT);</dd>
+          <dt><code>match_qualifier</code></dt>
+          <dd>A string to indicate how to match the values in <code>v</code>.
+            This can be used for general matching relations like "skos:exactMatch", "skos:closeMatch", etc. or for specific features like spatial matching with geo data
+            (e.g. containment search with "schema:containsPlace" etc.) or custom matching on date fields (e.g. services supporting the [[EDTF]] specification could use "EDTF:Level-0" etc.
+            To allow discovery of supported qualifiers by clients, services that support <code>match_qualifier</code> SHOULD return the supported <code>match_qualifiers</code> for each property
```
thadguidry, line 540: @fsteeg I think we're going to regret this where "match_qualifiers SHOULD return" instead of "MUST return".  But we'll see how things pan out over the next year.

fsteeg: I don't think this has been discussed before. I guess my motivation was to keep the barrier low for implementers, but using `MUST` sounds good to me too.

thadguidry: My thought was that discovery of supported qualifers will be impossible if we do not say MUST.
@wetneb your thoughts on this also?

wetneb: Yes, why not `MUST`, your reasoning makes sense to me.

fsteeg: Updated in https://github.com/reconciliation-api/specs/commit/1812f84d7f34a8342c0b7f2899ef032bcad6601a.

PR #133, opened 2023-07-13
Make reconciliation score optional

### LINKED ISSUE #127: Make scores of reconciliation candidates optional
During last month's meeting, we had an interesting discussion (with @paulgirard and @fsteeg) about the different strategies we can adopt around scoring. Broadly speaking, there are two approaches:
* adding more functionalities to the API so that clients can specify more precisely how the server should score the candidates (for instance how dates should be matched, which weights should be given to each properties, how to score multi-valued properties, how to match strings together, how to do type matching, and so on)
* exposing more information in the reconciliation responses so that the client can do a (re-)scoring on its own afterwards, to adjust the scoring to its needs. This means exposing some matching features, maybe property values, and so on.

Given that there are already some services which basically give up on returning a meaningful score for its reconciliation candidates (for instance by returning a constant score, or by returning one that is inversely proportional to the position of the candidate in the list), this suggests that we could make the `score` field of reconciliation candidates optional.

This would arguably be a cleaner option for those existing services. It would also give a clearer message to spec readers about this intention to make local re-scoring a more viable option and have less of a reliance on this server-side scoring.

wetneb:
Closes #127.

### DIFF
--- a/draft/examples/reconciliation-result-batch/valid/minimal.json
+++ b/draft/examples/reconciliation-result-batch/valid/minimal.json
@@ -1,0 +1,28 @@
+{
+  "results": [
+    {
+      "candidates": [
+        {
+          "id": "120333937",
+          "name": "Urbaniak, Regina"
+        },
+        {
+          "id": "1127147390",
+          "name": "Urbaniak, Jan"
+        }
+      ]
+    },
+    {
+      "candidates": [
+        {
+          "id": "123064325",
+          "name": "Schwanhold, Ernst"
+        },
+        {
+          "id": "116362988X",
+          "name": "Schwanhold, Nadine"
+        }
+      ]
+    }
+  ]
+}
--- a/draft/index.html
+++ b/draft/index.html
@@ -633,7 +633,7 @@ in the <code>score</code> field). By exposing individual features in their respo
 	</p>
 	<p>
 	  Deciding on a scoring method is one of the main difficulties in developing a reconciliation service.
-          Selecting candidates does not always rely on fuzzy matching. In those cases of exact match the score is an irrelevant information as candidates are either in or out. Thus the score attribute is optional.
+          Depending on the use case, we might not want a score at all: in cases like geo containment (a place should be inside a specific area) or date ranges (a point in time should be inside a specific time range) the score is irrelevant, as candidates are either in or out. Thus the score attribute is optional.
           Services are encouraged to expose as many <a>matching features</a> as they deem useful, in particular
           features which require knowledge of global statistics on the database or other attributes. Examples include:
           <dl>
--- a/draft/index.html
+++ b/draft/index.html
@@ -560,7 +560,7 @@ database are instances of this type.<dd>
 	    <dt><code>type</code></dt>
 	    <dd>The types of the candidate entity;</dd>
 	    <dt><code>score</code></dt>
-	    <dd>A numeral indicating how well this candidate entity matches the query: a higher score indicates a better match;</dt>
+	    <dd>An optional numeral indicating how well this candidate entity matches the query: a higher score indicates a better match;</dt>
             <dt><code>features</code></dt>
             <dd>An optional array of <a>matching features</a>;</dd>
 	    <dt><code>match</code></dt>
--- a/draft/index.html
+++ b/draft/index.html
@@ -560,7 +560,7 @@ database are instances of this type.<dd>
 	    <dt><code>type</code></dt>
 	    <dd>The types of the candidate entity;</dd>
 	    <dt><code>score</code></dt>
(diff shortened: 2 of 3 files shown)


draft/index.html
```diff
@@ -560,7 +560,7 @@ <h3>Reconciliation Query Responses</h3>
 	    <dt><code>type</code></dt>
 	    <dd>The types of the candidate entity;</dd>
 	    <dt><code>score</code></dt>
-	    <dd>A numeral indicating how well this candidate entity matches the query: a higher score indicates a better match;</dt>
+	    <dd>An optional numeral indicating how well this candidate entity matches the query: a higher score indicates a better match;</dt>
```
tfmorris, line 563: Is the list of reconciliation candidates rank ordered? I certainly think that would be most users' expectations. The spec seems silent on the issue which becomes much more noticeable in the absence of scores.

tfmorris (commented):
I don't have a strong objection to this, but it does seem like the spec is pulling in two different directions at once by introducing micro feature scores that a client needs to sum up / average while at the same time making the overall score optional.

As a user, I'm interested in easy to interpret, useful information to help efficiently review match candidates, including automatically accepting high confidence matches.

draft/index.html
paulgirard, line 563: Should the score be limited to a specific range in the spec ?
In the wikidata recon service scores are all bound to 0-100 range.
I don't have an opinion on this but it's the good place/time to discuss it.
The only input I can share is that having all the score on 0-100 range is clear and easy to interpret. But I don't know if some cases would need something different?

draft/index.html
paulgirard, line 1: I think the score being optional should be mentioned in the section "A Note on Candidate Retrieval and Scoring" https://github.com/reconciliation-api/specs/blob/b242f276bca50db3f136baf0091eb156f0cad641/draft/index.html#L623

Maybe something like 
```
Selecting candidates does not always rely on fuzzy matching. In those cases of exact match the score is an irrelevant information as candidates are either in or out. Thus the score attribute is optional.
```

paulgirard (changes requested):
I think that decision will clarify the candidate selection process. I proposed to add a sentence in the section about scoring. I added a proposal in a comment.

draft/index.html
tfmorris: Requiring normalization is tempting. The only other strong candidate for range would be 0-1.0.

Having said that, it opens a whole range of questions like:
- are scores comparable over time? for same search? for different searches? for different services?
- what does it mean to have a "perfect" score?

I think most search services punt on this and just say "bigger is better and you can't derive any additional information from scores," but, of course, users always try to infer some meaning anyway. Not sure what the best answer is...

paulgirard: We might want to explicitly state that the order has a sense only if score are provided.
A service not providing any score will provide an order which is totally arbitrary. 
So I propose to rephrase your proposition as

```
If candidates are scored, Reconciliation service must sort candidates in decreasing order of relevance i.e. be sorted in decreasing score order.
```

wetneb: I'd leave the score range unspecified for now, to avoid getting roped up into an independent consideration. We can introduce a normalization in a separate PR if it is desired.

draft/examples/reconciliation-result-batch/valid/minimal.json
```diff
@@ -0,0 +1,28 @@
+{
+  "results": [
+    {
+      "candidates": [
+        {
+          "id": "120333937",
+          "name": "Urbaniak, Regina"
+        },
+        {
+          "id": "1127147390",
```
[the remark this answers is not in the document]
wetneb: I cannot commit your suggestion because it is identical to the current state, apparently. ("This suggestion is invalid because no changes were made to the code.")

draft/index.html
```diff
@@ -633,6 +633,7 @@ <h3>A Note on Candidate Retrieval and Scoring</h3>
 	</p>
 	<p>
 	  Deciding on a scoring method is one of the main difficulties in developing a reconciliation service.
+          Selecting candidates does not always rely on fuzzy matching. In those cases of exact match the score is an irrelevant information as candidates are either in or out. Thus the score attribute is optional.
```
fsteeg, line 636: I feel like 'fuzzy matching' is confusing terminology here. E.g. a traditional search engine with basic TF-IDF scoring would not be considered 'fuzzy matching' in my opinion. I think it's a different level, e.g. when computing a score for a candidate, we can match individual strings with exact or fuzzy matching. To avoid that, maybe we could say something like this:

```suggestion
          Depending on the use case, we might not want a score at all: in cases like geo containment (a place should be inside a specific area) or date ranges (a point in time should be inside a specific time range) the score is irrelevant, as candidates are either in or out. Thus the score attribute is optional.
```

wetneb: I like your formulation, let's use it.

tfmorris (approved):

Issue #126, opened 2023-05-09
I18n checklist: text direction


fsteeg:
This is a self-review checklist based on the <a href="https://www.w3.org/TR/international-specs/#text_direction_checklist">Internationalization Best Practices for Spec Developers</a>.

(I've skipped what seemed formatting and markup related here.)

See also issue #52, #125 and notes at https://etherpad.lobid.org/p/reconciliation-i18n.

### Basic requirements
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> It must be possible to indicate base direction for each individual paragraph-level item of <a data-type="dfn" href="https://www.w3.org/TR/i18n-glossary/#dfn-natural-language">natural language</a> text that will be read by someone. <a href="https://www.w3.org/TR/international-specs/#dir_paragraphs" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> It must be possible to indicate base direction changes for embedded runs of inline bidirectional text for all <a data-type="dfn" href="https://www.w3.org/TR/i18n-glossary/#dfn-localizable-text">localizable text</a>. <a href="https://www.w3.org/TR/international-specs/#dir_inline" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_

### Background information
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> Do not assume that direction can be determined from language information. <a href="https://www.w3.org/TR/international-specs/#bidi_lang" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_

### Base direction values
1. [ ] <input type="checkbox" class="relevance" title="Check if this requirement is relevant to your spec."><input type="checkbox" class="ok" title="Check if your spec meets this requirement." onclick="this.previousSibling.checked=true"> Values for the default base direction should include left-to-right, right-to-left, and auto. <a href="https://www.w3.org/TR/international-specs/#bidi_values_req" class="checklistMore" target="_blank">more</a><br>    _Comments_go_here_

[...]

PR #137, opened 2023-09-13
Add section on setting the text direction

fsteeg:
For details see https://github.com/reconciliation-api/specs/issues/126 (where each item links to the relevant part of the W3C i18n docs).

### DIFF
--- a/draft/examples/data-extension-response/valid/example-dir.json
+++ b/draft/examples/data-extension-response/valid/example-dir.json
@@ -1,0 +1,37 @@
+{
+  "meta": [
+    {
+      "id": "Len",
+      "name": "Label (for language code 'en')"
+    },
+    {
+      "id": "Afa",
+      "name": "Also known as (for language code 'fa')"
+    }
+  ],
+  "rows": [
+    {
+      "id": "Q37093",
+      "dir": "ltr",
+      "properties": [
+        {
+          "id": "Len",
+          "values": [
+            {
+              "str": "Yahoo!"
+            }
+          ]
+        },
+        {
+          "id": "Afa",
+          "values": [
+            {
+              "str": "یاهو!",
+              "dir": "rtl"
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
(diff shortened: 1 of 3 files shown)


wetneb (commented):
Looks good to me. I wonder if we should add this field to the JSON schemas (and similarly the language field) or if it is just noise. Maybe it would be nice to have some examples use this (also to illustrate the inheritance described)?

fsteeg:
Yes, not sure about the schemas too, we have `lang` in a sample and it validates. For `dir` in a manifest: we currently have real-world sample manifests, so it would be weird to add it to an existing, or making one up. The manifest schema does not forbid `additionalProperties`, so manifests with `dir` validate. I don't think the schemas need to be something like a full spec of whats allowed in an element.

For the full example of `dir` and inheritance, I did try to come up with something similar to the `lang` example, but could not find anything useful in our GND data (from what I understand, setting the base direction  is required when mixing left-to-right and right-to-left strings). I tried constructing a data extension request against the Wikidata reconciliation endpoint (basically: give me the Persian label for 'Yahoo!'), but did not get that to work (the label itself is not a property? [1]). In the end I was happy with the little inline 'Yahoo!' example, in particular since the inheritance mechanism works just like `lang`, for which we have the full example.

[1] Tried smth. like this: https://wikidata.reconci.link/fa/api?extend={%22ids%22%3A[%22Q37093%22]%2C%22properties%22%3A[{%22id%22%3A%22P2561%22}]}

fsteeg:
With help from @wetneb in [today's meeting](https://etherpad.lobid.org/p/reconciliation-september-2023) I was able to come up with an example in [ea79f96](https://github.com/reconciliation-api/specs/pull/137/commits/ea79f966657b68010d8f9b77983c05bcddee958a) (based on a Wikidata [extension query]( https://wikidata.reconci.link/en/api?extend={%22ids%22%3A[%22Q37093%22]%2C%22properties%22%3A[{%22id%22%3A%22Afa%22}]}), which looks a bit different but would actually require setting a `dir` to render correctly).

PR #76, opened 2021-10-26
Adjust definition of schema space

acka47:
Resolves #3

PR #140, opened 2023-10-05
Remove identifierSpace and schemaSpace; make view templates mandatory

### LINKED ISSUE #139: Using identifierSpace when the entity set uses a mixture of namespaces
There has been some discussion on `identifierSpace` and `schemaSpace` before, e.g. in issue #3 and PR #76. The definitions of these have shifted over time. The [current definition](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#identifier-and-schema-spaces), in both the latest draft spec and version 0.2, of `identifierSpace` is:

> identifier space
>    The URI namespace (i.e. prefix) for the identifiers of an [entity](https://reconciliation-api.github.io/specs/draft/#dfn-entity) returned by the reconciliation service, for example http://www.wikidata.org/entity/ or https://d-nb.info/gnd/. This URI MAY resolve to a page describing these entities and their identifiers;

We are currently implementing reconciliation API support for Annif (see https://github.com/NatLibFi/Annif/pull/734) and providing the identifierSpace information has caused some headache. Returning the service manifest is mandatory, and also the identifierSpace information is mandatory within the manifest: "A reconciliation service MUST define two URIs [...] identifierSpace ... schemaSpace"

Service manifest [Example 1](https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/#example-1) given in the spec uses this identifierSpace:

> "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",

(FWIW, I would like to point out that this doesn't seem to match well with the definition - IIRC this is not the URI namespace prefix for any Getty vocabulary, but a URI/URL of a web page explaining them. But that is a separate problem, maybe the example is just outdated.)

Annif uses SKOS vocabularies internally and often those vocabularies use a specific URI namespace; in my understanding, this would be the natural value for identifierSpace. But Annif is currently unaware of this namespace, and there is nothing in principle preventing a vocabulary from using a mixture of namespaces. For example, a vocabulary could consist of a mixture of Wikidata and GND entities. A perhaps more realistic example would be a mixture of YSO concepts and those of a domain-specific extension vocabulary such as KAUNO (fiction literature), JUHO (public administration) or TERO (health and welfare), all of which are extensions of YSO - you can think of them naively as additional concepts to add on top of YSO - that use their own URI namespace which is different from YSO.

So what should Annif return in the service manifest for a project that uses a vocabulary whose URI namespace it isn't aware of? Should it look at all the concept URIs and try to infer what is the longest common prefix? What if the URIs are a mixture of namespaces and there is nothing in common - say, a mixture of http and https URIs?

Or should the value be something more custom (somewhat like the Getty document in the example) that isn't really a URI namespace at all, but is unique to the project? For example, the reconciliation service at /rest/v1/projects/myproject/reconcile could return an identifierSpace of /rest/v1/vocabs/myvocab (i.e. the vocabulary used by myproject). That doesn't seem to match the current definition of identifierSpace, as it talks specifically about URI namespace prefixes, but would at least be a shared identifier that could also be referenced by other endpoints at the same Annif instance which use the same underlying vocabulary.

Or is it OK to return an identifierSpace of `""` (the current quick-and-dirty solution in the Annif draft PR) since it seems to work fine with OpenRefine - apparently this information is not used at all. Maybe providing identifierSpace shouldn't be a MUST in the spec, if it's actually not used by the main client tool that this API is targeting.

osma:
This PR drops the concepts of identifierSpace and schemaSpace from the spec and instead makes view templates mandatory; this was suggested by @wetneb in https://github.com/reconciliation-api/specs/issues/139#issuecomment-1745125850.

Arguments for the change:

- The notions of identifierSpace and schemaSpace date from the Freebase era, as noted in the spec itself (the note is removed in this PR). Their relevance and usefulness in the current situation is at best unclear. Getting rid of cruft simplifies the spec and implementation of API services.
- Requiring the use of a single identifierSpace makes it impossible to provide entities from a mixture of URI namespaces. Dropping identifierSpace solves the problem. In a mixed namespace setting, a view template of `{+id}` can be used along with using full URIs as entity identifiers.
- View templates already provide a mechanism for turning entity identifiers into full URIs; this is more flexible than using identifierSpace. This PR makes the view templates mandatory, ensuring that entity identifiers can always be turned into URIs.
- The definition of schemaSpace has shifted; it was last changed for spec version 0.2 in PR #76, but it seems that current endpoints and manifest examples use schemaSpace in different ways (some use a type such as `skos:Concept`, while others link to ontologies or data model documentation).

Details of changes:

- Adjusted the spec itself
- Adjusted the schema
- Adjusted the examples

Closes #139

### DIFF
--- a/draft/examples/manifest/invalid/bad-authentication.json
+++ b/draft/examples/manifest/invalid/bad-authentication.json
@@ -1,7 +1,8 @@
 {
   "name": "OpenCorporates Reconciliation Service",
-  "identifierSpace": "http://rdf.freebase.com/ns/type.object.id",
-  "schemaSpace": "http://rdf.freebase.com/ns/type.object.id",
+  "view": {
+    "url": "{+id}"
+  },
   "authentication": {
      "foo": "bar"
   }
--- a/draft/examples/manifest/invalid/feature-view.json
+++ b/draft/examples/manifest/invalid/feature-view.json
@@ -1,7 +1,8 @@
 {
     "name": "VIAF",
-    "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",
-    "schemaSpace": "http://vocab.getty.edu/doc/#The_Getty_Vocabularies_and_LOD",
+    "view": {
+        "url": "https://viaf.org/viaf/{id}"
+    },
     "feature_view": {
         "url": "https://vocab.getty.edu/recon/features/"
     }
--- a/draft/examples/manifest/invalid/godot-missing-view.json
+++ b/draft/examples/manifest/invalid/godot-missing-view.json
@@ -8,8 +8,5 @@
   "name": "GODOT (Graph of Dated Objects and Texts) Reconciliation Service - Roman Consulates",
   "preview": {
     "url": "https://godot.date/reconcile/preview={id}"
-  },
-  "view": {
-    "url": "https://godot.date/id/{id}"
   }
 }
--- a/draft/examples/manifest/invalid/missing-version.json
+++ b/draft/examples/manifest/invalid/missing-version.json
@@ -1,5 +1,6 @@
 {
     "name": "VIAF",
-    "identifierSpace": "http://vocab.getty.edu/doc/#GVP_URLs_and_Prefixes",
-    "schemaSpace": "http://vocab.getty.edu/doc/#The_Getty_Vocabularies_and_LOD"
+    "view": {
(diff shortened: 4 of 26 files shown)


osma:
Rebased, fixed the conflict, force-pushed.

wetneb (approved):

fsteeg (approved):

PR #141, opened 2023-10-05
Change remaining URI templates in examples to RFC6570 syntax

osma:
While working on #140 , I noticed that some of the examples still use legacy view template syntax, i.e. `{{id}}` instead of `{id}` or `{+id}`. Some of the examples had been changed in #85 (that implemented #39), but not all of them. This fixes the remaining ones as well. I assume that this is what we want, to keep the examples in sync with the spec?


wetneb (approved):
Yes of course, thanks for catching that.

Issue #19, opened 2019-11-28
Do we really want to require JSONP support?


osma:
As currently written, the spec states that

> All HTTP(S) endpoints exposed by the service MUST support JSONP [RFC6902], which enables web-based clients to access the service from a different domain. 

Then a note says that OpenRefine 3.2 uses JSONP requests.

I wonder if this is a good idea going forward. JSONP was invented to circumvent CORS restrictions, but it has some problems (e.g. lack of error handling, excessive trust and potential CSRF vulnerabilities) - see [this blog post](https://blog.fullstacktraining.com/why-jsonp-shouldnt-be-used/) for some details. Nowadays better CORS techniques are well developed and supported in browsers.

I'm currently planning the implementation of the reconciliation API in Annif (see https://github.com/NatLibFi/Annif/issues/338) and having to implement JSONP looks like an extra obstacle. For one thing, Swagger/OpenAPI and the Connexion toolkit don't seem to support it that well - I may be mistaken here but e.g. the built in validation in Connexion probably won't play well with JSONP. These tools are really made for a post-JSONP world.

Of course I understand that OpenRefine itself may have to be changed if it currently still relies on JSONP support in the API endpoints. But I would suggest that the new spec should try to get rid of old baggage instead of cementing it. A first step would be to make CORS headers a MUST requirement (currently a SHOULD) and JSONP something less, perhaps a SHOULD.

wetneb:
I totally agree that moving to CORS is the way forward.

My intention with the current draft of the specs is that we document the protocol as it currently stands. I think this is a useful thing to have because many services have been designed according to those specs. All existing versions of OpenRefine (and its forks, rebranded versions…) only support JSONP at the moment, so if as a service provider you want to support these, it is useful to have this document available (as a prettier document than OpenRefine's wiki, basically). The intention is not to promote this as a "standard": it just documents how things stand.

Moving forward, I think we should make CORS a requirement ([and HTTPS too, probably](https://github.com/gbv/nomisma-reconciliation)) but but only in a new version of the specs, such that this is properly recorded as a change (since it is a breaking one).

We can totally start working on a new version of the specs which would include this: we just need to figure out how to handle versioning: #7.

osma:
Thanks for the quick response @wetneb! I understand the strategy of first documenting the current API and only then making a new, potentially incompatible version. Perhaps ideas for spec changes could have a specific issue label in GitHub, to keep them separate from housekeeping tasks such as most of the currently open issues.

[...]

PR #142, opened 2023-10-05
Remove JSONP support

osma:
While I'm at it making PRs, I propose that we drop JSONP support altogether. This PR does it.

As already explained in #19, it's an obsolete an insecure mechanism for making cross-origin HTTP requests. It was already made optional for spec version 0.2, but I think it's better to leave it out from the API spec entirely because CORS is already mandated (since 0.2) and does the same thing, but in a better way.

### DIFF
--- a/draft/index.html
+++ b/draft/index.html
@@ -61,11 +61,6 @@
         localBiblio: {
           "beek-2018": {"aliasOf": "doi:10.1007/978-3-319-93417-4_5"},
           "christen-2012": {"aliasOf": "doi:10.1007/978-3-642-31164-2"},
-          "JSONP": {
-            "title": "JSONP",
-            "publisher": "English Wikipedia",
-            "href": "https://en.wikipedia.org/wiki/JSONP"
-          },
           "CSRF": {
             "title": "Cross-Site Request Forgery",
             "publisher": "English Wikipedia",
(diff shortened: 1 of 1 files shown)


wetneb (approved):
Yes, JSONP was only there for backwards compatibility, but since this new spec has breaking changes it does not make sense to keep it. Services are free to still implement version 0.1 or 0.2 of the specs on top.

osma:
Thanks for the quick approval @wetneb!
Can I just merge this or do we need to wait for opinions from others?

wetneb:
It seems relatively uncontentious but let's wait for another opinion, I guess it should not have merge conflicts with other open PRs anyway.

osma:
The problem with merge conflicts is caused by the list of changes included in the spec. I thought it would make sense to add a brief mention of the changes I made in the same PR that implements those changes. But that inevitably leads to merge conflicts when there are more than one open PRs making that kind of changes. Fortunately those should be easy to resolve.

draft/index.html
```diff
@@ -222,7 +217,8 @@ <h4>This Draft</h4>
           <ul>
             <li>Remove support for flyouts;</li>
             <li>The endpoints of various sub-services are no longer configurable, but determined by the root reconciliation endpoint;</li>
-            <li>The reconciliation queries only support passing a single <code>type</code> and the <code>type_strict</code> parameter was removed.</li>
+            <li>The reconciliation queries only support passing a single <code>type</code> and the <code>type_strict</code> parameter was removed;</li>
+            <li>Remove JSONP support.</li>
```
tfmorris, line 221:
```suggestion
            <li>The reconciliation queries only support passing a single <code>type</code> and the <code>type_strict</code> parameter was removed</li>
            <li>Remove JSONP support</li>
```
I'm not sure if there's a specific style guide being followed which requires this punctuation, but it seems odd to me. It's certainly not commonly used in American English. I would suggest removing all the terminal punctuation for the list entries. (Applies to the two list entries above as well)

tfmorris (changes requested):
Looks good except for a minor nit with list punctuation (which wasn't introduced in this PR, but made noticeable because of it).

draft/index.html
osma: Yeah, I noticed that as well, but I figured that it's best to just keep following the established style.

Should I remove the punctuation in this PR? Make another PR? There are two more lists that use similar punctuation above this one (the ones for versions 0.1 and 0.2).

wetneb: Agreed, we can change the formatting of the list in another PR.

PR #149, opened 2023-12-14
Move 'query' field inside 'properties'

### LINKED ISSUE #106: Reformulate the `query` parameter as one of the `properties` in reconciliation queries

### LINKED ISSUE #134: Submitting multiple entity names in the same reconciliation query
When reconciling entities, it is not uncommon that I know multiple names for the entities I am looking for. For instance, a full name and an abbreviated one, or names in different languages, and so on.

At the moment, the reconciliation protocol does not provide a simple way to provide all of those names together in the same reconciliation query.

One possible workaround as a service is to offer a matching property which behaves like the supplied name, such that clients can send properties such as
```json
{
    "query": "A name for the entity",
    "properties": [
        { "pid": "name", "v": "Another name for the entity" }
    ]
}
```

The problem with this approach is that the `"pid": "name"` part is not standard: clients (and their users) need to discover those properties by themselves. For instance, in the Wikidata reconciliation endpoint, the pid which plays this role is `Len|Aen` (for labels and aliases in English) - good luck with discovering this! And it's not exactly equivalent to supplying a value in `"query"` since that matches against labels and aliases in all languages.

### Specify a standard property id for extra names

This would make it possible for clients (such as OpenRefine) to offer the matching of additional names in a clearer way.

Maybe the `query` field could be removed in that case, so that all values are supplied via the `properties` array. See #106.

### Make the `query` field accept arrays

Reconciliation queries of the form
```
{
     "query": [ "A name for the entity", "Another name for the entity" ]
}
```
could be submitted. Maybe queries with a single name should be reformulated to wrap the value in an array anyway, for consistency.

### Any other?

### LINKED ISSUE #145: Reformulate the 'type' parameter as another property
Given that we are thinking about changing the `query` field of reconciliation queries into a property (https://github.com/reconciliation-api/specs/issues/134), I think it's tempting to think whether we could do the same for types. It would then let us use the additional settings on properties (https://github.com/reconciliation-api/specs/pull/131), essentially offering users the choice between:
* union of types (`any`)
* intersection of types (`all`)
* complement of the union (`none`)

One problem is that the `property values` we currently allow in a reconciliation query do not include type ids. I am not too sure what to do about this.

But overall I feel like #115 is likely going in the wrong direction, given the recurrent feedback we get in OpenRefine about needing support for reconciling against multiple types. (See the discussion in #109 too)

wetneb:
Closes #134. Closes #106.


draft/index.html
```diff
@@ -478,29 +478,35 @@ <h2>Reconciliation Queries</h2>
...
-		How reconciliation services handle this further restriction ("must match all properties" or "should match some") and how it affects the score, is up to the service.
-		A reconciliation service that supports properties SHOULD provide a <a>suggest service</a> for discovering these properties;</dd>
+            <dd>An optional limit on the number of candidates to return, which must be a positive integer;</dd>
           </dl>
         </p>
+        <p>
+          A <dfn>property assignment</dfn> specifies the expected value of a property on the entities to match.
+                These are used to filter the set of candidates (similar to a WHERE clause in SQL),
+		by allowing clients to specify an attribute of entities that should match. It consists of:
+         <dl>
```
osma, line 498: HTML nitpick: Isn't `<dl>` a block element in HTML? It shouldn't be placed inside `<p>` like this, right?

osma (approved):
From my perspective, this is a somewhat unexpected solution for #134 and #145. In a way I like it, because it unifies the structure of queries. OTOH, it makes the simple case of querying just for a name more complex, as seen from the examples; the minimal example is a lot less minimal now.

The PR looks well made (except for a minor HTML structure issue that I commented on) and seems to correctly adjust all the relevant examples.

I'm not against this if the CG decides that this is the right way forward. Just a bit concerned about the additional complexity.

draft/index.html
wetneb: Good point, I created an issue about it because it would require normalization in the rest of the document too: https://github.com/reconciliation-api/specs/issues/151

draft/examples/reconciliation-query-batch/invalid/misnamed-property.json
```diff
@@ -1,10 +1,12 @@
 {
   "queries": [
     {
-      "query": "Christel Hanewinckel",
       "type": "DifferentiatedPerson",
       "limit": 5,
       "props": [
+        {
+          "v": "Christel Hanewinckel"
```
tfmorris, line 8: Why no property? Shouldn't it be `rdfs:label` or something similar?

draft/index.html
```diff
@@ -481,42 +481,44 @@ <h2>Reconciliation Queries</h2>
...
+        </p>
+        <p>
+          A <dfn>property assignment</dfn> specifies the expected value of a property on the entities to match.
+                These are used to filter the set of candidates (similar to a WHERE clause in SQL),
+		by allowing clients to specify an attribute of entities that should match. It consists of:
+         <dl>
+            <dt><code>pid</code></dt>
+            <dd>A <a href='#properties'>property</a> identifier. If this is not provided, then this signals that
+                the client intends to search for entities with similar names. The specifics of how this similarity
+                is defined are determined by the service.</dd>
```
tfmorris, line 505: Having a missing property signal a label/name search seems potentially error-prone.

draft/schemas/reconciliation-query-batch.json
```diff
@@ -61,13 +57,14 @@
...
-            "description": "An optional list of property mappings to refine the query",
+            "minItems": 1,
+            "description": "A list of property mappings to select candidates",
             "items": {
               "type": "object",
               "properties": {
                 "pid": {
                   "type": "string",
-                  "description": "The identifier of the property, whose values will be compared to the values supplied"
+                  "description": "The identifier of the property, whose values will be compared to the values supplied. If absent, values will be matched against the entity names"
```
tfmorris, line 67: Same comment as above.

draft/examples/reconciliation-query-batch/invalid/multiple-types.json
```diff
@@ -1,8 +1,15 @@
 {
   "queries": [
     {
-      "query": "Christel Hanewinckel",
-      "type": ["DifferentiatedPerson", "FictionalCharacter"],
+      "properties": [
```
tfmorris, line 4: The examples above use `props` not `properties`. Are both accepted or is one of them wrong?

tfmorris (commented):
If an incompatible change is going to be introduced in the name of orderliness, I think it makes sense to make it as clean as possible and tidy up all the loose ends. In particular, having a missing property imply some special property doesn't seem very clean.

draft/index.html
thadguidry: Why would it generate errors?  A default fallback strategy for "Things not Strings" would indeed be if Things (properties) are not provided then a String (name/label) is searched.  And so we accounted for that with query or properties must be supplied in the query syntax.  In the current draft:

> 4.1 Structure of a Reconciliation Query #structure-of-a-reconciliation-query
>A **reconciliation query** consists of the following fields. At least one of `query` or `properties` must be supplied, but all other fields are optional.
>
>**query**
>A query string, consisting of a non-empty string. By supplying such a string, a client intends to search for entities with similar names. The specifics of how this similarity is defined are determined by the service.

1. So maybe we just need to reword and cleanup 503-505 to align with what we already say at the top of the spec in `#structure-of-a-reconciliation-query`
2. capitalize MUST in the query structure wording - At least one of `query` or `properties` MUST be supplied, but all other fields are optional.

draft/examples/reconciliation-query-batch/invalid/multiple-types.json
thadguidry: I don't think we should accept both.  `properties` should be the only one, in my opinion to avoid confusion.

wetneb: Only "properties" is accepted. If you are referring to [draft/examples/reconciliation-query-batch/invalid/misnamed-property.json](https://github.com/reconciliation-api/specs/pull/149/files/399d613ac0d6eb7b647e500121d5c494c15ed0b9#diff-902a970609dca55e649281df9fb19e1fdb52c2520b70d222702bf0813fc872c5), then it is intentional that this JSON is wrong (see the `invalid/misnamed-property` part of the path - it is a negative example in our test suite). If you are aware of any other place where we use `props`, it's likely a problem that needs fixing indeed.

draft/examples/reconciliation-query-batch/invalid/misnamed-property.json
wetneb: Let's keep the discussion in one place, below

draft/index.html
wetneb: In our monthly call we weighed the pros and cons of having a pre-defined id, mandated by the specs, versus other solutions and the group that was on the call converged towards the solution of not including a pid. The main reluctance I have towards using an id mandated by the specs is that it can conflict with the naming choices of other properties in the service. By using `rdfs:label` specifically I'd be a bit worried that we are sort of implying that this has something to do with RDF, which it does not (so far at least). To me it seems to be mostly a matter of taste, honestly. I am happy to adapt the PR to another consensus if there is one.

osma: @thadguidry 

>  But it was removed in this commit and now seems kinda important to state that somehow. Seems like we need to be explicit and mention if `query` is always required (MUST be supplied) inside properties. I don't see this clearly, and previously we had the anyOf.

I wonder if there's a misunderstanding here. This PR removes `query` as a special JSON keyword altogether. It just becomes another `v` entry under `properties`, with an empty `pid` to signal that it is expected to match names or label-like properties. On line 486, it is stated that `properties` is "An array of property assignments (at least one)." which is equivalent to the previous anyOf statement.

thadguidry: @osma Misunderstanding indeed.  Sorry, and why I stopped commenting when I realized what was happening. :-)  I've made a note in our etherpad to maybe bring clarity further https://etherpad.lobid.org/p/reconciliation-january-2024#L32  I'll be on the meeting call later today.

tfmorris: If that was the consensus of the expert group, I'm fine with it. The "error prone"ness that I was talking about was for developers who accidently forget to include an ID, not production scenarios, but they'll figure it out.

draft/examples/reconciliation-query-batch/invalid/multiple-types.json
tfmorris: Sorry, I missed that was an error case.

wetneb:
@tfmorris in our call this week, we thought that we could accommodate your preference for an explicit `pid`, because we have an eye on https://github.com/reconciliation-api/specs/issues/145 as well, for which we would likely need a similar solution (and we would not be able to use an empty pid for both, obviously).

Also, I want to stress that the consensus we reach in the monthly calls are not (in my opinion) meant to be definitive resolutions of the entire community group, but rather ways to quickly run ideas by other group members and get their feedback early on (saving some rounds of review on GitHub). So there is no problem with questioning those outcomes. That's why we still make PRs for those things.

osma:
I think it was me who suggested an empty pid, in the spirit of "what is the simplest thing that could possibly work?". I have no problem admitting that this is hackish and ugly.

> we thought that we could accommodate your preference for an explicit pid, because we have an eye on https://github.com/reconciliation-api/specs/issues/145 as well, for which we would likely need a similar solution (and we would not be able to use an empty pid for both, obviously).

It seems to me that the scope of `properties` is expanding, perhaps way beyond its original intent, if we are going to apply it not just to labels but to types as well. How about renaming it into something more generic such as `conditions` (or `where`) and adding a new attribute `match_type` to specify whether we are matching labels, properties or types? It could look something like this:

```json
      "conditions": [
        {
          "match_type": "label",
          "v": "Christel Hanewinckel"
        },
        {
          "match_type": "property",
          "v": "Politik*",
          "required": false,
          "match_quantifier": "any",
          "match_qualifier": "WildcardMatch"
        },
        {
          "match_type": "property",
          "pid": "affiliation",
          "v": "http://d-nb.info/gnd/2022139-3",
          "required": false,
          "match_quantifier": "any",
          "match_qualifier": "ExactMatch"
        }
      ]
``` 

To handle types as well (#145), we could add `"match_type": "type"` as well.

fsteeg:
> To handle types as well (https://github.com/reconciliation-api/specs/issues/145), we could add `"match_type": "type"` as well.

So to summarize, we'd have 3 different kinds of `match_type` for these proposed `conditions`: `label` (replacing the old `query`, no `pid` required), `type` (replacing the old `type`, no `pid` required), and `property` (the original case, `pid` required).

For compatibility, we could also keep the `properties` name, define `"match_type": "property"` as the default, and effectively only add a new optional field `match_type`? And for the possible values we could stick with the existing `query`, `type`, `property` as values.

It seems like a sensible way to structure this. In general, simply using `label` and `type` as `pid` and not adding a new field seems simpler. However, with a `match_type` we'd avoid possible collisions with existing `label` or `type` fields (and workarounds like `_label` / `_type`), right? And maybe it's actually cleaner to have a separate field, instead of giving certain `pid` values a special meaning?

wetneb:
That sounds good to me, but I would suggest `"match_type": "name"` instead of `"match_type": "label"` since that's currently how we call this field in the specs (and JSON serialization of reconciliation candidates, for instance).

draft/index.html
```diff
@@ -481,42 +481,46 @@ <h2>Reconciliation Queries</h2>
...
+            <dd>An optional limit on the number of candidates to return, which must be a positive integer;</dd>
+          </dl>
+        </p>
+        <p>
+          A <dfn>condition</dfn> specifies a constraint that should be matched by the entities to return.
+                It is used to filter the set of candidates (similar to a WHERE clause in SQL),
+		by allowing clients to specify an attribute of entities that should match. It consists of:
+         <dl>
+            <dt><code>match_type</code></dt>
+            <dd>Either <code>name</code> or <code>property</code>, depending on whether the condition related to entity names or their properties.</dd>
```
fsteeg, line 503:
```suggestion
            <dd>Either <code>name</code> or <code>property</code>, depending on whether the condition relates to entity names or their properties.</dd>
```

osma:
Should the changes made in this PR also be mentioned in the changes section [1.4.3 This Draft](https://reconciliation-api.github.io/specs/draft/#this-draft)?

wetneb:
Yes, we could add update this section with all of the recent changes we have done (not just this one).

PR #156, opened 2024-03-14
Add optional `standardizedScore` field in service manifest (#147)

### LINKED ISSUE #147: Advertise score range used by a service in its manifest
Services are free to use any numerical range for the scores they give to reconciliation candidates.

As a client, this makes it a bit hard to process those scores: we can only compare the candidates between each other and have no idea what the maximum or minimum score might be.

Services could advertise their min and max scores in their manifest (if any). Some might use unbounded scores on purpose and it could be nice to be able to indicate that too.

Alternatively, the specifications could mandate a particular score range.

Reconciliation features could also have a similar metadata.

fsteeg:
Will resolve #147.

### DIFF
--- a/draft/examples/manifest/valid/wikidata.json
+++ b/draft/examples/manifest/valid/wikidata.json
@@ -82,5 +82,6 @@
     "type": true,
     "entity": true,
     "property": true
-  }
+  },
+  "standardizedScore": true
 }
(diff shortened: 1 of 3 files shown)


wetneb (approved):
Simple and does the job. I like it!

fsteeg:
As mentioned by @osma in https://github.com/reconciliation-api/specs/pull/149#issuecomment-1997665231 for #149, I should also update the [This Draft](https://reconciliation-api.github.io/specs/draft/#this-draft) section here.

fsteeg:
Added to "This Draft" section, merging as discussed in [today's meeting](https://etherpad.lobid.org/p/reconciliation-april-2024).

PR #163, opened 2024-04-06
Fix inconsistent data extension metadata example

wetneb:
The associated schema (in the manifest) is already up to date.
That's not caught by our CI because we don't have a schema only for data extension metadata, but for the whole manifest instead.

Follow-up to #91.


fsteeg (approved):
(Also touched in [#166](https://github.com/reconciliation-api/specs/pull/166/files#diff-ee829a2bd7db186d382cc947c69272c3d3f0a5d870e80dfd7d58e8bdf81acdb9), should conflict.)

PR #169, opened 2024-04-11
Clarify data type as integer for `limit` parameter

### LINKED ISSUE #167: Should the data type for `limit` parameter be a String or Number?
The current wording for `limit` under Data Extension Property Proposals says
> Optionally, the requested limit

and ideally would instead give the data type

> An optional string value to control the number of proposed properties fetched.

We do have a paragraph further above that `limit` parameter that mentions it's a String data type? as 

> The service SHOULD support an optional limit query string parameter to control the number of proposed properties.

But I noticed that we have other numerical values for things like this, such as Query's `limit`, and Preview's `width`, `height` and also Candidates `score`, and wondered if we should ensure that we use numerical data types for parameter values always expected to be a number like `limit` ?

thadguidry:
Fixes #167
- Add data type 'integer' to the description for `limit`.
- Also removes "string" in "query string parameter" which made it read and sound like the parameters' were sometimes String datatype. I hate English sometimes.

### DIFF
--- a/draft/index.html
+++ b/draft/index.html
@@ -847,7 +847,7 @@ containing a type identifier.
             <dt><code>type</code></dt>
             <dd>The <a>type</a> identifier supplied in the query;</dd>
             <dt><code>limit</code></dt>
-            <dd>Optionally, an integer to control the number of proposed properties returned for the given type;</dd>
+            <dd>Optionally, an integer to indicate the requested limit;</dd>
           </dl>
         </p>
         <p>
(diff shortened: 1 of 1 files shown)


draft/index.html
```diff
@@ -847,7 +847,7 @@ <h3>Data Extension Property Proposals</h3>
             <dt><code>type</code></dt>
             <dd>The <a>type</a> identifier supplied in the query;</dd>
             <dt><code>limit</code></dt>
-            <dd>Optionally, the requested limit;</dd>
+            <dd>Optionally, an integer to control the number of proposed properties returned for the given type;</dd>
```
fsteeg, line 850: Since this is the response, not the request, I'd suggest something like this:

```suggestion
            <dd>Optionally, an integer to indicate the requested limit;</dd>
```

wetneb (approved):

PR #173, opened 2024-04-14
Begin prototyping data types of fields and optional

thadguidry:
- I'm not 100% sure if we should link using ecmascript, or another standard for the data types.  WhatWG has Infra https://infra.spec.whatwg.org/#string as an example, but then I didn't see Array.  Still it would be nice to link and point to a standard, but that would be JSON, and I couldn't figure out how to make the ReSpec shortcut links work with RFC8259 and so resorted to the shortcut links using ecmascript (but this all could be changed I'm sure if we knew more or had a friend in ReSpec land)

### DIFF
--- a/draft/index.html
+++ b/draft/index.html
@@ -5,6 +5,7 @@
     <script src='../libs/respec-w3c.js' async class='remove'></script>
     <script class='remove'>
       var respecConfig = {
+        xref: ["ecmascript", "RFC2119", "RFC6570"],
         specStatus: "CG-DRAFT",
         subtitle: "A protocol for data matching on the Web",
         pluralize: true,
@@ -244,7 +245,6 @@ href="https://github.com/OpenRefine/OpenRefine/wiki/Reconciliable-Data-Sources">
           </ul>
         </section>
       </section>
-      <section id="conformance"></section>
     </section>
     <section>
       <h2>Core Concepts</h2>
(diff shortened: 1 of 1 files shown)


draft/index.html
```diff
@@ -255,21 +255,53 @@ <h2>Core Concepts</h2>
...
+        <b>JSON</b> [[RFC8259]] format conventions are used throughout this document, including examples.
+      </p>  
+      <p>
+        The following concepts are used and borrowed from JSON syntax as well as from [[RFC2119]]:
+      <ul>
+        <li>
+        <dfn
+          data-lt="key|keyword|fields">
+          field
+        </dfn> ,plural "fields", is the alternative term used throughout to refer to the key or name part in a <code>"name":"value"</code> pair in a JSON object as defined by [[RFC8259]].
```
fsteeg, line 269:
```suggestion
        </dfn> is the alternative term used throughout to refer to the key or name part in a <code>"name":"value"</code> pair in a JSON object as defined by [[RFC8259]].
```

fsteeg (commented):
Formatting looks good, gives nice overview of data type & optional vs. required.

Made one inline suggestion. Also ReSpec complains about the unused `required` definition. I think omitting `required` looks good though, so maybe we can just remove the `dfn` and make it bold instead? Or link to it once from some place?

thadguidry:
> Formatting looks good, gives nice overview of data type & optional vs. required.

Thanks!

> Made one inline suggestion. Also ReSpec complains about the unused `required` definition. I think omitting `required` looks good though, so maybe we can just remove the `dfn` and make it bold instead? Or link to it once from some place?

Yeah, agree.  So, instead, I now have just put a comment inside doc about that, and then made it only bold/italic to match.
Also wrapped paratheses around them to showcase the syntax used in the document.  I think this makes things easier to distinguish.

Finally, does anyone have opinions on the linking for the data types?  Notice, if you hover or click on "String" or "Array" etc. that it's currently linked to "ecmascript".  Is that OK you think?  or should we link to some other standard available in [xref](https://respec.org/docs/#xref) ?

thadguidry:
I'll instead link the Data Types against the [RFC8259 JSON Specification ](https://www.rfc-editor.org/rfc/rfc8259)

draft/index.html
```diff
@@ -255,21 +255,71 @@ <h2>Core Concepts</h2>
...
-           </dl>
+        <dl>
+        <dt><code>id</code></dt>
+          <dd>{{String}} a non-empty identifier. This identifier must be unique among all entities;</dd>
+        <dt><code>name</code></dt>
+          <dd>{{String}} ({{optional}}) a non-empty <emph>name</emph> or <emph></emph>label</emph> for the entity;</dd>
+        <dt><code>description</code></dt>
+          <dd>{{String}} ({{optional}}) a <emph>description</emph> as a human-readable string;</dd>
+        <dt><code>type</code></dt>
+          <dd>{{Array}} ({{optional}}) list of <a>types</a>, possibly empty, the entity is classified by;</dd>
```
thadguidry, line 321: @fsteeg Does this appended phrase make things more clear, or does it assume too much?  I felt like it was missing this context, which might not be apparent to developers outside of Semantic Web or Knowledge Graph circles (or just getting introduced to them) when we're talking about Entities and their Types.

thadguidry:
@fsteeg
1. So I layered in the JSON spec conventions (and removed the ecmascript xref since we agreed to conform to just JSON spec), and then with those defined, I now began to add more datatypes.  How do things look now (links look/feel good) (format is just fine)?
2. Also took the liberty to slightly clarify a few things and improve grammar. One thing I am wondering about is clarifying the `Entities` - type, where I appended `... the entity is classified by`.  Does that make good sense?

fsteeg:
Looks good to me in principle. Seeing the type directly makes it clearer. It feels a bit like there's a delimiter (e.g. a dot?) missing between the type and the description, but maybe it's fine. I think the description should be consistently capitalized though.

thadguidry:
@fsteeg Let's just take this as an example to come to consensus, and to ensure I understand what you are asking for:

`<dd>{{Array}} ({{optional}}) list of <a>types</a>, possibly empty, the entity is classified by;</dd>`

1. You want a dot between the datatype (or optional) and the description?  Such as:
`<dd>{{Array}} ({{optional}}). List of <a>types</a>, possibly empty, the entity is classified by;</dd>`
2. So you think descriptions should always start with a captial such as "list" should be "List" ?

fsteeg:
Yes, that's what I meant.

wetneb (commented):
I'm not super fond of redefining the JSON concepts. Intuitively we should be able to just link to whatever spec defines that and avoid making our spec even longer (I think it's already a bit scary because of its length).

Concerning the formatting of field types, it feels to me that we are missing a sort of delimiter between the field type and its description. An alternative way of formatting this would be to have the field name, type, optional status and description all gathered in a table.

wetneb:
Nice, I think it works to have `String (optional)` followed by the description of the field, it's quite natural.

PR #176, opened 2024-07-10
Change candidate `name` and `description` from string to array of objects with required `str`, optional `lang` and `dir` fields

### LINKED ISSUE #138: Support for multi-lingual candidate names
As a service provider, I would like clients to be able to query in any language and to return candidate names in one or more  languages specified by the client request.

Use Case
=======
A client is reconciling a place in Canada using the Artsdata.ca Reconciliation service with the name "Studio Azrieli". 

Current solution (not ideal)
--------------------
The service returns multiple entities including  [K11-15](http://kg.artsdata.ca/resource/K11-15) "National Arts Centre - Azrieli Studio" and [K11-15](http://kg.artsdata.ca/resource/K11-15) "Centre National des Arts - Studio Azrieli" which appear as separate entities but have the same URI. This may appear incorrect to the user because there are 2 candidates. If the user doesn't notice that they have the same URI then they may be mistaken as duplicates.

Screenshot 2023-09-14 at 9 52 11 AM

Ideal solution
--------------
The service returns multiple entities but only a single [K11-15](http://kg.artsdata.ca/resource/K11-15) displaying both names "National Arts Centre - Azrieli Studio" and "Centre National des Arts - Studio Azrieli" together.  Parameters can specify the languages the client would like to display.

fsteeg:
To resolve #138.

See also https://etherpad.lobid.org/p/reconciliation-july-2024.

### DIFF
--- a/draft/examples/reconciliation-candidate/valid/example.json
+++ b/draft/examples/reconciliation-candidate/valid/example.json
@@ -1,6 +1,10 @@
 {
   "id": "1117582299",
-  "name": "Urbaniak, Hans-Eberhard",
+  "name": [
+    {
+      "str": "Urbaniak, Hans-Eberhard"
+    }
+  ],
   "score": 85.71888,
   "features": [
     {
--- a/draft/examples/reconciliation-result-batch/invalid/missing-id.json
+++ b/draft/examples/reconciliation-result-batch/invalid/missing-id.json
@@ -3,7 +3,11 @@
     {
       "candidates": [
         {
-          "name": "Urbaniak, Regina",
+          "name": [
+            {
+              "str": "Urbaniak, Regina"
+            }
+          ],
           "score": 53.015232,
           "match": false,
           "type": [
--- a/draft/examples/reconciliation-result-batch/invalid/root-array.json
+++ b/draft/examples/reconciliation-result-batch/invalid/root-array.json
@@ -3,7 +3,11 @@
     [
       {
         "id": "120333937",
-        "name": "Urbaniak, Regina",
+        "name": [
+          {
+            "str": "Urbaniak, Regina"
+          }
+        ],
         "score": 53.015232,
         "match": false,
         "type": [
@@ -19,7 +23,11 @@
       },
       {
         "id": "1127147390",
-        "name": "Urbaniak, Jan",
+        "name": [
+          {
+            "str": "Urbaniak, Jan"
+          }
+        ],
         "score": 52.357353,
         "match": false,
         "type": [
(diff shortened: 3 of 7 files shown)


wetneb (commented):
Very nice!
I wonder if the same syntax should be adopted elsewhere, for instance in the responses for entity, name and type preview. I think it would be more consistent since those can be localized too, no?

wetneb (commented):
Very nice!
I wonder if we should adopt the same format for names and descriptions in other places in JSON serializations, such as in suggest responses. I think that would be more consistent.

fsteeg:
> I wonder if the same syntax should be adopted elsewhere, for instance in the responses for entity, name and type preview

You mean the suggest services (entity, property, type)? Additionally there are names and descriptions in data extension property proposals, names in data extension responses and a service name in the manifest. If we aim for a consistent data structure, should these all use the array-of-objects syntax?

At the same time, we support both simple strings and arrays of objects (or strings, even mixed) as `condition` values (`v`), see https://reconciliation-api.github.io/specs/draft/#example-7:

```json
"conditions": [
    {
      "matchType": "name",
      "v": "Christel Hanewinckel"
    },
    {
      "matchType": "property",
      "v": [
        "Politik*",
        {
          "id": "wissenschaftler",
          "name": "Wissenschaftler(in)"
        }
      ]
    }
]
```

For consistency and maintainability we should probably unify these, right? To keep the common case simple, I'd prefer the current `v` behavior for all strings, but the discussion in #138 very much favors the consistent array-of-objects syntax (though it mostly mentions only the array part). But then `v` should also require consistent array-of-object values, right?

wetneb:
I think I have a preference for keeping the same format all the time (`[{"str":"some value"}]` in minimal form) even if it's a bit more verbose, because otherwise we're bound to have implementations which oversee this possibility and break when confronted with something else than a bare string.

fsteeg:
I've updated suggest responses and property proposals to use the string object format too.

While there are fields all over the spec that in theory could also use the string object format, I think restricting it to the current state in this PR does actually make some sense: candidates, suggest responses and property proposals are all provided to users for making a selection. So this would allow multi-language objects for all user selections. At the same time, I keep wondering if this actually is the right approach. In his original requirement in #138, @saumier wrote:

> [...] to return candidate names in one or more languages specified by the client request

Did we ever consider implementing it by (only) specifying the language(s) in the client request (in the Accept-Language header), and returning the old structure? Do we actually need to return multiple languages at the same time, for a single request?

I also noticed that in many places we don't use the actual object structure of our concepts (like entities, properties, types), but objects representing these concepts, but adding context-specific fields. I tried to clarify that a bit for the places I switched to the string object format.

PR #177, opened 2024-09-10
Specify defaults for `required` and `matchQuantifier` (#155)

### LINKED ISSUE #155: Clarify default value of 'required' and 'match_quantifier' field on matching conditions
In https://reconciliation-api.github.io/specs/draft/#structure-of-a-reconciliation-query, we currently have:

> required
>     A boolean indicating if a match for the property is required for an entity to enter the list of candidates (i.e. acting like a filter or a WHERE clause in SQL) or optional (i.e. only effecting the entity's rank in the list of candidates);

It would be useful to clarify the implied value of this parameter (if any) when it is not present, and similarly for `match_quantifier`.

fsteeg:
In a reconciliation query's condition object

Will resolve #155.

### DIFF
--- a/draft/index.html
+++ b/draft/index.html
@@ -523,18 +523,18 @@ database are instances of this type.<dd>
                 is defined are determined by the service.</dd> 
             <dt><code>required</code></dt>
             <dd>An optional boolean indicating if a match for the property is required for an entity to enter the list of candidates (i.e. acting like a filter or a WHERE clause in SQL)
-              or optional (i.e. only effecting the entity's rank in the list of candidates);</dd>
+              or optional (i.e. only effecting the entity's rank in the list of candidates). If omitted, the service SHOULD assume the requested value to be <code>false</code>;</dd>
             <dt><code>matchQuantifier</code></dt>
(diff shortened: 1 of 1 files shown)


draft/index.html
```diff
@@ -523,18 +523,18 @@ <h3>Structure of a Reconciliation Query</h3>
...
             <dt><code>matchQualifier</code></dt>
             <dd>An optional string to indicate how to match the values in <code>v</code>.
               This can be used for general matching relations like "skos:exactMatch", "skos:closeMatch", etc. or for specific features like spatial matching with geo data
               (e.g. containment search with "schema:containsPlace" etc.) or custom matching on date fields (e.g. services supporting the [[EDTF]] specification could use "EDTF:Level-0" etc.
               To allow discovery of supported qualifiers by clients, services that support <code>matchQualifier</code> MUST return the supported <code>matchQualifiers</code> for each property
-              in their property <a href='#suggest-responses'>suggest responses</a>.</p></dd>
+              in their property <a href='#suggest-responses'>suggest responses</a>. If omitted, the exact matching behavior is up to the service.</p></dd>
```
thadguidry, line 534:
```suggestion
              in their property <a href='#suggest-responses'>suggest responses</a>. If omitted, the matching behavior is up to the service.</p></dd>
```
Do we really need the "exact" adverb here?  Probably not?

thadguidry (changes requested):

wetneb (approved):

PR #179, opened 2024-09-12
Set version to 1.0-draft

wetneb:
Implementations of the current draft struggle to announce themselves properly because the specs don't specify how they should describe their version in the `versions` field of the manifest.

This:
* adjusts the specs so that the `1.0-draft` version is specified as such a value
* also changes the URL (with a redirection) so that the version is clear from there
* updates the JSON schema to require this value to be present
* updates the JSON examples accordingly

The choice of `1.0-draft` was the consensus that arose from today's meeting, considering that
* we have made quite a lot of breaking changes since 0.2 so it's worth signifying that
* we don't intend to make much more versions than this so it feels appropriate to aim for a 1.0 soon


1.0-draft/index.html
```diff
@@ -0,0 +1,1138 @@
...
+        pluralize: true,
+	doJsonLd: true,
+        group: 'reconciliation',
+        wg: "Entity Reconciliation Community Group",
+        wgPublicList: "public-reconciliation",
+        wgURI: "https://www.w3.org/community/reconciliation/",
+	canonicalURI: "https://reconciliation-api.github.io/specs/draft/",
+	edDraftURI: "https://reconciliation-api.github.io/specs/draft/",
+        prevRecURI: "https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/",
+        latestVersion: "https://www.w3.org/community/reports/reconciliation/CG-FINAL-specs-0.2-20230410/",
```
thadguidry, line 19: Will these need to be updated as well?

1.0-draft/index.html
```diff
@@ -0,0 +1,1138 @@
...
+        <h3>Service Manifest</h3>
+        <p>
+           When the reconciliation service <a>endpoint</a> is queried with a HTTP GET query without
+           parameters, the <a>service manifest</a> MUST be returned.
+        </p>
+      <p>
+        A <dfn>service manifest</dfn> consists of the following fields:
+        <dl>
+          <dt><code>versions</code></dt>
+          <dd>The array of API versions supported by the endpoint, such as <code>["0.1", "0.2"]</code>. Since this field did not exist in version 0.1, services which do not declare a <code>versions</code> field are expected to only support version 0.1. A service implementing the current draft should include <code>1.0-draft</code> in the array.</dd>
```
thadguidry, line 357:
```suggestion
          <dd>The array of API versions supported by the endpoint, such as <code>["0.1", "0.2", "1.0-draft"]</code>. Since this field did not exist in version 0.1, services which do not declare a <code>versions</code> field are expected to only support version 0.1. A service implementing the current draft should include <code>1.0-draft</code> in the array.</dd>
```

Probably best to make the array complete, for now.

1.0-draft/index.html
wetneb: Good catch!

thadguidry (approved):

draft/index.html
```diff
@@ -1,1138 +1,15 @@
...
-        <p>
-          The <a href="schemas/data-extension-response.json">data extension response schema</a> validates <a>data extension responses</a>.
-        </p>
-        <pre data-include="schemas/data-extension-response.json" />
-      </section>
-    </section>
-  </body>
+	<head>
+		<meta charset="utf-8" />
+		<meta http-equiv="Refresh" content="0; url='https://reconciliation-api.github.io/specs/0.2/'" />
```
fsteeg, line 5:
```suggestion
		<meta http-equiv="Refresh" content="0; url='https://reconciliation-api.github.io/specs/1.0-draft/'" />
```

PR #195, opened 2025-10-08
Updated spec and schema to add "id" matchType

saumier:
Updated spec and schema to add "id" as an explicit matchType option for reconciliation queries.


fsteeg (commented):
Looks good, just added two minor suggestions for consistent punctuation.

wetneb (approved):

PR #200, opened 2026-07-15
Section 3.7 Version Negotiation. closes #78

### LINKED ISSUE #78: Negotiating API versions
It looks like the API version here doesn't change much, but I'm not clear what the semantics are when the API does change.  As mentioned [here](https://github.com/reconciliation-api/specs/pull/67#issuecomment-968246070) the [service manifest](https://reconciliation-api.github.io/specs/latest/#service-manifest) includes a list of versions the server can support, but it's not clear what that means.

Is the client expected, to pick from among those versions?  Does the client fail when it can't support any of those versions?  If the server supports multiple versions, how does the client communicate that to the server?  What exactly is the handshake mechanism?

Again, this doesn't seem urgent as the API version doesn't change much but having a well defined plan could help in figuring out what risks there are to a given change and mapping out a rollout of added features.

saumier:


1.0-draft/index.html
```diff
@@ -519,6 +519,84 @@ <h3>Authentication</h3>
...
+        <p>
+          To let a service support multiple protocol versions on the same endpoint, clients MAY express the version they expect
+          in the HTTP <code>Accept</code> header with a reconciliation-specific media type.
+        </p>
+        <p>
+          A client requesting version 1 can send:
+          <pre class="example nohighlight">Accept: application/reconciliation.v1+json</pre>
+        </p>
+        <p>
+          A service that supports multiple versions SHOULD inspect this header and serve the highest-priority compatible version.
```
fsteeg, line 533: To actually support multiple versions, to me it seem the service MUST check the header:

```suggestion
          A service that declares support for multiple versions in its manifest MUST inspect this header and serve the highest-priority compatible version.
```

1.0-draft/index.html
```diff
@@ -519,6 +519,84 @@ <h3>Authentication</h3>
...
+        <p>
+          Version matching for reconciliation-specific media types SHOULD be exact: for instance,
+          <code>application/reconciliation.v1+json</code> and <code>application/reconciliation.v1.1+json</code> identify
+          different protocol variants.
+        </p>
+        <p>
+          Services SHOULD only apply fallback when no reconciliation-specific media type is supplied in <code>Accept</code>.
+          In that case, services SHOULD use a documented default version for unversioned requests,
+          such as their latest stable version.
+          If a request explicitly advertises only unsupported reconciliation versions, services SHOULD return
```
fsteeg, line 547: I find 'advertises' a bit confusing here. How about this:


```suggestion
          If a client explicitly requests only unsupported reconciliation versions, services SHOULD return
```

1.0-draft/index.html
```diff
@@ -519,6 +519,84 @@ <h3>Authentication</h3>
...
+          to identify the semantic profile of the selected representation.
+          This <code>Link</code> value is sent in the HTTP response headers.
+        </p>
+        <p>
+          For debugging, services MAY also include an informational header such as:
+          <code>Reconciliation-Version: 1.0</code>
+          If provided, this header is informational only and MUST NOT override the negotiated
+          representation indicated by <code>Content-Type</code>.
+        </p>
+
```
fsteeg, line 578: Isn't this redundant? We already provide that info in the Content-Type and the Link headers, plus it's the only non-standard header here, and it adds ambiguity (which you clarify in the text). So from my point of view, this could be removed:


```suggestion
```

fsteeg (commented):
Thanks Gregory! Looks good overall, some inline suggestions below.

1.0-draft/index.html
```diff
@@ -519,6 +519,84 @@ <h3>Authentication</h3>
...
+          in the HTTP <code>Accept</code> header with a reconciliation-specific media type.
+        </p>
+        <p>
+          A client requesting version 1 can send:
+          <pre class="example nohighlight">Accept: application/reconciliation.v1+json</pre>
+        </p>
+        <p>
+          A service that supports multiple versions SHOULD inspect this header and serve the highest-priority compatible version.
+          Reconciliation-specific media types in <code>Accept</code> SHOULD take precedence over generic media types such as
+          <code>application/json</code>. Clients MAY use standard HTTP quality factors in <code>Accept</code> to express preference
```
wetneb, line 535: Nice, I didn't know about this mechanism. I propose to link to the specs directly, for the reader's convenience:
```suggestion
          <code>application/json</code>. Clients MAY use standard HTTP quality factors as defined in [[RFC9110]] to express preference
```

1.0-draft/index.html
wetneb: I tend to agree, we can just stick to the normative part.

1.0-draft/index.html
```diff
@@ -519,6 +519,84 @@ <h3>Authentication</h3>
...
+        </p>
+        
+        <p>
+          Example of a successful negotiated response header:
+          <pre class="example nohighlight">HTTP/1.1 200 OK
+Date: Wed, 15 Jul 2026 14:22:31 GMT
+Content-Type: application/reconciliation.v1+json
+Vary: Accept
+Link: &lt;https://reconciliation-api.github.io/specs/1.0/&gt;; rel="profile"
+Reconciliation-Version: 1.0
```
wetneb, line 586: Per the above, I'd remove this header line.

wetneb (commented):
Thanks for working on this! I'm going to look into implementing this for the Wikibase recon service, with version 0.2 as the default (meaning, only offer support for 1.0 for clients which pass this `Accept` header).

